PR #57, opened 2026-07-28
feat(storage): implement CSEK for uploads

### LINKED ISSUE #25: [storage upload]: uploads should support CSEK

chingor13:
Fixes #25


packages/storage/Sources/GoogleCloudStorage/UploadOptions.swift
```diff
@@ -86,15 +105,93 @@ public struct ChecksumOptions: Sendable, Hashable {
...
+  }
+
+  public var keyHashBase64: String {
+    let data = key.withUnsafeBytes { Data($0) }
+    let hash = SHA256.hash(data: data)
+    return Data(hash).base64EncodedString()
+  }
+
+  public var description: String {
+    "CustomerEncryptionKeyOptions(algorithm: \(algorithm), keyHashBase64: \(keyHashBase64))"
```
coryan: Nice... we are not leaking the secret in the logs!

packages/storage/Sources/GoogleCloudStorage/UploadOptions.swift
```diff
@@ -86,15 +105,93 @@ public struct ChecksumOptions: Sendable, Hashable {
...
 /// [standard Base64]: https://datatracker.ietf.org/doc/html/rfc4648#section-4
 /// [Customer-Supplied Encryption Keys]: https://docs.cloud.google.com/storage/docs/encryption/customer-supplied-keys
-public struct CustomerEncryptionKey: Sendable {
+public struct CustomerEncryptionKeyOptions: Sendable, Hashable, Equatable, CustomStringConvertible,
+  CustomDebugStringConvertible
+{
   public let algorithm: String
-  public let keyBase64: String
-  public let keyHashBase64: String
+  public let key: SymmetricKey
```
coryan, line 135: Does it still make sense to keep this public?

packages/storage/Sources/GoogleCloudStorage/UploadOptions.swift
```diff
@@ -76,6 +77,24 @@ public struct ChecksumOptions: Sendable, Hashable {
   }
 }
 
+/// Errors thrown when validating or creating a `CustomerEncryptionKeyOptions`.
+public enum CustomerEncryptionKeyError: Error, Sendable, Hashable, Equatable,
```
coryan: Does this need to be `Hashable` ? I know it is trivial to implement, but once we promise a public type conforms to a protocol, we can never take that away. Better to make as few promises as possible.

packages/storage/Sources/GoogleCloudStorage/UploadOptions.swift
```diff
@@ -86,15 +105,93 @@ public struct ChecksumOptions: Sendable, Hashable {
...
-public struct CustomerEncryptionKey: Sendable {
+public struct CustomerEncryptionKeyOptions: Sendable, Hashable, Equatable, CustomStringConvertible,
+  CustomDebugStringConvertible
+{
   public let algorithm: String
-  public let keyBase64: String
-  public let keyHashBase64: String
+  public let key: SymmetricKey
+
+  public func hash(into hasher: inout Hasher) {
```
coryan: Ditto on making this type `Hashable`... Unless we are planning to put them in hash tables or hash sets, we should avoid promising more than we need.

packages/storage/Sources/GoogleCloudStorage/UploadOptions.swift
```diff
@@ -86,15 +105,93 @@ public struct ChecksumOptions: Sendable, Hashable {
...
+    self.algorithm = algorithm
+    self.key = SymmetricKey(data: keyData)
+  }
+
+  /// Creates a `CustomerEncryptionKeyOptions` from raw key bytes (`Data`).
+  ///
+  /// For the default "AES256" algorithm, the key must be exactly 32 bytes (256 bits).
+  /// The key and its SHA-256 hash are automatically Base64-encoded.
+  public init(key: Data, algorithm: String = "AES256") throws {
+    if algorithm == "AES256" && key.count != 32 {
```
coryan: Seems like "AES256" appears often enough that it deserves a constant?

packages/storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
```diff
@@ -303,6 +304,24 @@ extension HTTPClient {
     var queryItems = [URLQueryItem(name: "uploadType", value: "multipart")]
     queryItems.append(URLQueryItem(name: "name", value: objectName))
 
+    if let kmsKeyName = options.kmsKeyName {
+      queryItems.append(URLQueryItem(name: "kmsKeyName", value: kmsKeyName))
+    }
+    if let preconditions = options.preconditions {
+      if let ifGen = preconditions.ifGenerationMatch {
```
coryan: I believe each `if let ` creates its own scope so you could say:


```suggestion
      if let v = preconditions.ifGenerationMatch {
```

packages/storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
```diff
@@ -341,17 +366,44 @@ extension HTTPClient {
...
+      }
+      if let ifGenNot = preconditions.ifGenerationNotMatch {
+        queryItems.append(URLQueryItem(name: "ifGenerationNotMatch", value: String(ifGenNot)))
+      }
+      if let ifMeta = preconditions.ifMetagenerationMatch {
+        queryItems.append(URLQueryItem(name: "ifMetagenerationMatch", value: String(ifMeta)))
+      }
+      if let ifMetaNot = preconditions.ifMetagenerationNotMatch {
+        queryItems.append(URLQueryItem(name: "ifMetagenerationNotMatch", value: String(ifMetaNot)))
+      }
```
coryan: This seems like a strict dup of the code above, can we refactor it to a private function or something?

packages/storage/Tests/CustomerEncryptionKeyTests.swift
```diff
@@ -0,0 +1,257 @@
...
+import Crypto
+import Foundation
+#if canImport(FoundationNetworking)
+  import FoundationNetworking
+#endif
+import GoogleCloudGax
+@testable import GoogleCloudStorage
+import Testing
+
+@Suite struct CustomerEncryptionKeyOptionsTests {
```
coryan, line 24: This seems to include tests for the `CustomerEncryptionKeyOptions` type and for how the storage client uses that type. Am I right? If so, do you mind splitting them?

packages/storage/Sources/GoogleCloudStorage/UploadOptions.swift
```diff
@@ -86,15 +105,93 @@ public struct ChecksumOptions: Sendable, Hashable {
...
+    -> Bool
+  {
+    lhs.algorithm == rhs.algorithm && lhs.key == rhs.key
+  }
+
+  public var keyBase64: String {
+    key.withUnsafeBytes { Data($0).base64EncodedString() }
+  }
+
+  public var keyHashBase64: String {
```
coryan, line 149: If these are really going to be public properties, they need documentation.

coryan (changes requested):
This is pretty good. A lot of nits about conforming to more protocols than we need. Please consider the nits and remove the code duplication (if it is real, I may be missing something).

packages/storage/Sources/GoogleCloudStorage/UploadOptions.swift
chingor13: I was also considering turning it into a constant

packages/storage/Sources/GoogleCloudStorage/UploadOptions.swift
chingor13: Removed `Hashable`

packages/storage/Sources/GoogleCloudStorage/UploadOptions.swift
chingor13: All the other options types are currently public. If we want to restrict visibility, we should probably do it for all the options types.

packages/storage/Sources/GoogleCloudStorage/UploadOptions.swift
chingor13: Removed

packages/storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
chingor13: Refactored into a private helper on the preconditions

packages/storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
chingor13: Refactored into a shared helper

packages/storage/Tests/CustomerEncryptionKeyTests.swift
chingor13: split the class tests from the HTTP mock tests

packages/storage/Sources/GoogleCloudStorage/UploadOptions.swift
chingor13: Added

coryan (approved):

PR #60, opened 2026-07-28
chore!: rename libraries used as deps

coryan:
In this pass, I renamed all the libraries that are used as dependencies of other libraries (e.g. `GoogleIdentityAccesscontextManagerV1`) to use `PascalCase` for each word.

That affects a lot of the generated libraries, as `GoogleIAMV1` is used in many of them.

Towards googleapis/librarian#6229


coryan (commented):
Please double check `librarian.yaml` for any spelling errors.

librarian.yaml
```diff
@@ -44,40 +44,40 @@ default:
       - name: GoogleCloudGax
         path: packages/gax
         required_by_services: true
-      - name: GoogleCloudGkehubConfigmanagementV1
+      - name: GoogleCloudGkeHubConfigManagementV1
```
chingor13: Should GkeHub be GKEHub?

I think for this section, it doesn't really matter, but later in the library_name_override it might?

coryan: > Should GkeHub be GKEHub?

Good question...  I think you are correct. We should use `GKE` consistently.  The inner types may be wrong, but nothing we can do about those.

> I think for this section, it doesn't really matter, but later in the library_name_override it might?

We need both to match.

coryan (commented):
PTAL

chingor13 (approved):

PR #64, opened 2026-07-28
impl(Gax): use polling policies in LROs

### LINKED ISSUE #61: Use the polling policies `GoogleCloudGax._PollableOperationImpl`

coryan:
Change the LRO polling loop to use a polling policy. The policy can stop the iteration if the polling loop is exhausted or the error is not considered retryable.

Fixes #61

### DIFF
--- a/packages/gax/Sources/GoogleCloudGax/LongRunningOperations.swift
+++ b/packages/gax/Sources/GoogleCloudGax/LongRunningOperations.swift
@@ -60,7 +60,8 @@ public final class _PollableOperationImpl<ResponseType>: PollableOperation {
   private var state: State
   private let pollOp: Poll
   private let sleep: Sleep
-  private let backoffPolicy: BackoffPolicy = LinearBackoffPolicy()
+  private var backoffPolicy: BackoffPolicy = LinearBackoffPolicy()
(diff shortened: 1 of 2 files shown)


chingor13 (approved):

packages/gax/Sources/GoogleCloudGax/LongRunningOperations.swift
```diff
@@ -88,11 +96,24 @@ public final class _PollableOperationImpl<ResponseType>: PollableOperation {
   ///   - The underlying operation error if the operation failed.
   ///   - `RequestError.malformedResponse` if the operation completed successfully but returned no result.
   public func wait() async throws -> ResponseType {
-    let retryState = RetryState.init()
+    var pollingState = PollingState()
     while !state.done {
-      let delay = backoffPolicy.backoffDelay(for: retryState)
+      try pollingPolicy.onInProgress(state: pollingState, name: "")
```
chingor13, line 101: The empty name arg does look strange here.

coryan: You are right, that parameter will not work. I should remove it. I will do so in a follow up PR.

PR #65, opened 2026-07-28
chore!: rename more libraries

coryan:
In this pass, I renamed any libraries where capitalizing each word in the package name produced an obviously bad name in `PascalCase` style.

Several other libraries will need overrides: other SDKs have chosen to rename things like `GoogleDevToolsTraceV2` to `GoogleCloudTraceV2`. And some SDKs have opinions about `Bigquery` vs. `BigQuery`. I think it will be easier to grok such changes if they come in separate PRs.

Continues the work from googleapis/librarian#6229


chingor13 (changes requested):
Can you rewrite the commits to have a single commit that has the librarian.yaml changes? As is, we can't review this PR effectively from the UI

coryan:
> Can you rewrite the commits to have a single commit that has the librarian.yaml changes? As is, we can't review this PR effectively from the UI

Done. I should have thought of that beforehand.

chingor13 (approved):

PR #66, opened 2026-07-28
chore!: rename Bigquery libraries

coryan:
Override the name for all Bigquery libraries. Several needed the override to make the name readable. In a few Bigquery libraries I overrode the name to be consistent and silence the errors from future `librarian` versions.

Towards googleapis/librarian#6229


coryan:
I will resend this after other changes.

PR #77, opened 2026-07-30
test(storage): use anonymous credentials

### LINKED ISSUE #3: Get the storage unit tests to pass on GHA
They seem to be using the GCE metadata server:

```
✘ Test resumeLocalSourceTooSmall() recorded an issue at UrlMocks.swift:97:17: Issue recorded
  ↳ Expected error of type UploadError, but got .missingEnvironmentConfiguration("MDS environment probe failed for target URL: http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token. GCE_METADATA_HOST environment variable not found. Found environment keys: []. Underlying error: transportError(Error Domain=NSURLErrorDomain Code=-1003 \"(null)\"UserInfo={NSLocalizedDescription=Could not resolve host: metadata.google.internal, NSErrorFailingURLKey=http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token, NSUnderlyingError=Error Domain=NSURLErrorDomain Code=-1003 \"(null)\"UserInfo={NSLocalizedDescription=Could not resolve host: metadata.google.internal}, NSErrorFailingURLStringKey=[http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token})](http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token%7D))")
```

coryan:
By default, the storage client (like all clients) use MDS credentials. These are not available in the test environment, so the tests failed after being unable to initialize the request.

With this change, the tests use anonymous credentials, which always work.

Fixes #3

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -40,8 +40,6 @@ flags=(
 )
 for dir in "${packages[@]}"; do
     [[ -f "${dir}/Package.swift" ]] || continue
-    # TODO(https://github.com/googleapis/google-cloud-swift/issues/3) - restore these builds.
-    [[ "${dir}" != "packages/storage" ]] || continue
     count=$((count + 1))
 
     echo "::group::--- Building ${dir} ---"
--- a/packages/storage/Tests/ChecksumTests.swift
+++ b/packages/storage/Tests/ChecksumTests.swift
@@ -16,6 +16,7 @@ import Foundation
 #if canImport(FoundationNetworking)
   import FoundationNetworking
 #endif
+import GoogleCloudAuth
 import GoogleCloudGax
 @testable import GoogleCloudStorage
 import Testing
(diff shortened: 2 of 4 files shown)


chingor13 (changes requested):
Is this actually a bug in our ADC implementation? Should we be falling back to anonymous credentials if we can't detect credentials from the runtime environment?

chingor13:
> Is this actually a bug in our ADC implementation? Should we be falling back to anonymous credentials if we can't detect credentials from the runtime environment?

I guess other languages are throwing errors if ADC cannot be determined

chingor13 (approved):

coryan:
> Is this actually a bug in our ADC implementation? Should we be falling back to anonymous credentials if we can't detect credentials from the runtime environment?

The fallback is to error out. I hate how ADC is described, but...

> 2.3 Credentials not found [END]
> 3. Check workload credentials (on GCE, GKE, GAE and Serverless)
> ...
> 3.2 If false, go to step (2.3)

https://google.aip.dev/auth/4110

So you could argue that the test should fail because the credentials are not found... but I don't think we can fallback on anonymous.

Issue #12, opened 2026-07-25
Fix documentation warnings in hand-crafted libraries


coryan:
Seems like we made some mistakes in cross-referencing elements and then did not fix them.

- [x] Fix documentation warnings in `packages/wkt`
- [x] Fix documentation warnings in `packages/gax`
- [ ] Create CI that generates docs for `packages/*` with `--warnings-as-errors`.

coryan:
This is now done for realsies. The build exists too.

PR #78, opened 2026-07-30
ci: bootstrap GCB configuration

coryan:
We will need to use GCB (Google Cloud Build) for larger CI builds and for any integration tests against production. This PR bootstraps the necessary resource management using Terraform.

The README files explain why I prefer using Terraform to manage the CI resources.

Keeping the Terraform configuration in the repo allows us to evolve the configuration at the same time as the project.

Towards #12 and several other CI tasks

### DIFF
--- a/.gcloudignore
+++ b/.gcloudignore
@@ -1,0 +1,18 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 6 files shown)


chingor13:
So this is terraform to set up a long-lived bucket that will be used to hold terraform state for temporary resources (using a different terraform config?)

coryan:
> So this is terraform to set up a long-lived bucket that will be used to hold terraform state for temporary resources (using a different terraform config?)

For temporary and long running resources.... It is a drag that you cannot use the same terraform to initialize a bucket for its state and then use that bucket to store the rest of the state.... At least I don't know of a way.

chingor13 (approved):

PR #106, opened 2026-07-31
fix(storage): calculate checksum automatically on resuming a failed upload

### LINKED ISSUE #105: [storage upload] resuming upload does not include automatic checksumming
Currently, we seek to the resuming offset, however, we do not ever re-read the bytes to do the calculation.

chingor13:
Fixes #105 

This refactors when we wrap the upload source with `ChecksummedSource` and whether or not we can call `seek` on the `ChecksummedSource`. When we call `seek` on `ChecksummedSource` we are now re-calculating the on-the-fly checksumming by re-reading from the start of the source.


packages/storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
```diff
@@ -238,14 +283,11 @@ extension StorageClient {
...
 
-      // 2. Seek source to GCS offset
-      try await source.seek(to: offset)
-
       continuation.yield(
         UploadStatus(
           bytesUploaded: offset, totalBytes: totalSize, uploadId: uploadId))
 
-      // 3. Continue upload
+      // 2. Continue upload
```
coryan: You are about to call the `continueResumableUpload` method. This comment is kind of superfluous?

packages/storage/Sources/GoogleCloudStorage/ChecksummedSource.swift
```diff
@@ -106,3 +106,46 @@ struct ChecksummedSource<S: UploadSource> {
...
+    let needMD5 = (options.md5 == .auto)
+
+    guard needCRC32C || needMD5 else {
+      // no checksums requested, nothing more to do
+      try await source.seek(to: offset)
+      return
+    }
+
+    guard offset > 0 else {
+      // we are starting at the beginning of the file, no precompuation needed
```
coryan:
```suggestion
      // we are starting at the beginning of the file, no precomputation needed
```

packages/storage/Sources/GoogleCloudStorage/ChecksummedSource.swift
```diff
@@ -106,3 +106,46 @@ struct ChecksummedSource<S: UploadSource> {
...
+      try await source.seek(to: 0)
+      return
+    }
+
+    // Start at the beginning of the file to compute checksum up to offset
+    try await source.seek(to: 0)
```
coryan: I may be misunderstanding, when I had to deal with this problem before the checksummed source (or its moral equivalent) only updated the hash / checksum if the bytes were "new"... No need to re-read the old bytes: on the first pass your update the hash / checksum, on the second pass you ignore the old data, and only start updating on new bytes.  There is a corner case when source.read() gives you partially new data, but it is not an impossible problem.

There is also the case where the data jumps ahead... but then caller violated an invariant anyway.

coryan (changes requested):
I may be misunderstanding the intent, but re-reading the data from the beginning seems sad, consider a large upload (multiple GiBs) that needs to rewind just to resend the last few bytes: are you going to re-read all the data again just to update the checksums?

packages/storage/Sources/GoogleCloudStorage/ChecksummedSource.swift
chingor13: You're right, we can optimize this better if we still have the previously calculated values still in memory. If we don't then we will have to re-compute from the beginning. I was thinking about a scenario where we have to start fresh because the process crashed or something.

coryan: I think we should optimize for the common case, which are (AFAIU):

- Everything uploads without problems on the first attempt 
- We need to resume, but the data is still in memory (10x or maybe 10x less common than the previous bullet point)
- We need to resume, and the client has crashed (maybe 10x less common than the previous bullet point)

But maybe the new query information is there and we can stop worrying?

packages/storage/Sources/GoogleCloudStorage/ChecksummedSource.swift
```diff
@@ -49,6 +49,13 @@ struct ChecksummedSource<S: UploadSource> {
     }
   }
 
+  mutating func seedCRC32C(seed: UInt32, bytesHashed: Int64) {
+    if options.crc32c == .auto {
+      self.crc32c = CRC32C(seed: seed)
```
coryan: You probably need to rebase, I moved the `CRC32C` class to `gax` and renamed it to `_CRC32C` to make it clear it is an internal-only thing.

packages/storage/Sources/GoogleCloudStorage/ChecksummedSource.swift
chingor13: This logic is now updated:

1. If we have the same upload source instance, we maintain a pointer for how far we've read/checksummed. We will continue checksumming after we read new bytes.
2. If we don't have the same upload source instance, the server responds with the running crc32c hash when we ask for the upload status. We use this value to seed our crc32c calculation and continue checksumming from there

coryan (approved):

PR #112, opened 2026-08-03
feat(storage): regenerate conversions for optional, nested fields

nolanleastin:
Regenerate StorageControl conversion extensions to support optional presence checks, nested message fields, Well-Known Types (Timestamp), and GoogleType.Interval.

- Configured google/type module output path in librarian.yaml.
- Regenerated all StorageControl conversion extensions under Sources/GoogleCloudStorage/generated/.
- Added unit tests in ConversionTests.swift covering GetFolderRequest optional presence, Folder nested message & WKT mappings, and GoogleType.Interval roundtrip conversions.

### DIFF
--- a/packages/storage/Sources/GoogleCloudStorage/generated/StorageControl/Convert/GoogleType/Interval+Convert.swift
+++ b/packages/storage/Sources/GoogleCloudStorage/generated/StorageControl/Convert/GoogleType/Interval+Convert.swift
@@ -27,8 +27,8 @@ extension Interval {
 
   internal init(proto: ProtoType) throws {
     self.init()
-    self.startTime = proto.hasStartTime ? try GoogleCloudWkt.Timestamp(proto: proto.startTime) : nil
-    self.endTime = proto.hasEndTime ? try GoogleCloudWkt.Timestamp(proto: proto.endTime) : nil
+    self.startTime = proto.hasStartTime ? try .init(proto: proto.startTime) : nil
+    self.endTime = proto.hasEndTime ? try .init(proto: proto.endTime) : nil
   }
 
   internal func toProto() throws -> ProtoType {
--- a/packages/storage/Sources/GoogleCloudStorage/generated/StorageControl/Convert/GoogleType/Interval+Convert.swift
+++ b/packages/storage/Sources/GoogleCloudStorage/generated/StorageControl/Convert/GoogleType/Interval+Convert.swift
@@ -19,28 +19,22 @@ import GoogleCloudGax
 internal import StorageControlProtos
 internal import SwiftProtobuf
 import GoogleCloudWkt
+import GoogleType
 internal import GoogleCloudWktConvert
 
-extension UpdateOrganizationIntelligenceConfigRequest {
-  internal typealias ProtoType = StorageControlProtos
-    .Google_Storage_Control_V2_UpdateOrganizationIntelligenceConfigRequest
+extension Interval {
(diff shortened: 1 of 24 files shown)


packages/storage/Sources/GoogleCloudStorage/generated/StorageControl/Convert/GoogleType/Interval+Convert.swift
```diff
@@ -0,0 +1,40 @@
...
+import GoogleCloudWkt
+import GoogleType
+internal import GoogleCloudWktConvert
+
+extension Interval {
+  internal typealias ProtoType = StorageControlProtos.Google_Type_Interval
+
+  internal init(proto: ProtoType) throws {
+    self.init()
+    self.startTime = proto.hasStartTime ? try GoogleCloudWkt.Timestamp(proto: proto.startTime) : nil
```
coryan: I wonder if you can get away with:


```suggestion
    self.startTime = proto.hasStartTime ? try .init(proto: proto.startTime) : nil
```

I just don't know if Swift's type deduction is strong enough.

packages/storage/Sources/GoogleCloudStorage/generated/StorageControl/Convert/GoogleType/Interval+Convert.swift
```diff
@@ -0,0 +1,40 @@
...
+    self.endTime = proto.hasEndTime ? try GoogleCloudWkt.Timestamp(proto: proto.endTime) : nil
+  }
+
+  internal func toProto() throws -> ProtoType {
+    var proto = ProtoType()
+    if let startTime = self.startTime { proto.startTime = try startTime.toProto() }
```
coryan, line 36: Why not map?


```suggestion
    proto.startTime = self.startTime.map { try $0.toProto() }
```

packages/storage/Sources/GoogleCloudStorage/generated/StorageControl/Convert/StorageControl/FindingSummary+Convert.swift
```diff
@@ -26,12 +26,78 @@ extension FindingSummary {
 
   internal init(proto: ProtoType) throws {
     self.init()
+    self.type = FindingType(proto: proto.type)
```
coryan: Pretty sure you can use 

```suggestion
    self.type = .init(proto: proto.type)
```

packages/storage/Sources/GoogleCloudStorage/generated/StorageControl/Convert/GoogleType/Interval+Convert.swift
nolanleastin: oh i din't know about that, it worked!

packages/storage/Sources/GoogleCloudStorage/generated/StorageControl/Convert/GoogleType/Interval+Convert.swift
nolanleastin: Responded in the librarian PR, but `.map { try $0.toProto() }` evaluates to `Google_Protobuf_Timestamp?` (Optional) .... which causes a Swift compiler error when assigned:

```
proto.startTime: error: cannot assign value of type 'Google_Protobuf_Timestamp?' to type 'Google_Protobuf_Timestamp'
```

coryan (approved):

PR #113, opened 2026-08-03
fix(storage): set content-range headers correctly for streaming sources of unknown length

### LINKED ISSUE #26: [storage upload]: supported upload sources
- [ ] fixed buffers in memory
- [ ] files from disk
- [ ] dynamically created via computation
- [ ] data received asynchronously from external source (e.g. download)

chingor13:
Also adds tests for all our supported upload source types

Fixes #26

### DIFF
--- a/packages/storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
+++ b/packages/storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
@@ -153,18 +153,24 @@ extension StorageClient {
     continuation: AsyncStream<UploadStatus>.Continuation
   ) async throws -> StorageObject {
     var offset = offset
+    var chunksSent = 0
     while true {
       guard let chunkInfo = try await checksummedSource.readChunk(maxBytes: chunkSize),
         !chunkInfo.data.isEmpty
       else {
         break
       }
+      chunksSent += 1
       let chunk = chunkInfo.data
       let isLast = chunkInfo.isLast
       let checksum = isLast ? chunkInfo.checksum : nil
 
+      let effectiveTotalSize =
+        (isLast && totalSize == nil) ? (offset + Int64(chunk.count)) : totalSize
+
       let uploadRequest = try await httpClient.buildUploadChunkRequest(
-        uploadId: uploadId, data: chunk, offset: offset, totalSize: totalSize, options: options,
+        uploadId: uploadId, data: chunk, offset: offset, totalSize: effectiveTotalSize,
+        options: options,
         checksum: checksum)
       let (uploadData, uploadResponse) = try await httpClient.data(for: uploadRequest)
 
@@ -174,7 +180,7 @@ extension StorageClient {
         continuation.yield(
           UploadStatus(
             bytesUploaded: offset + Int64(chunk.count),
-            totalBytes: totalSize, uploadId: uploadId))
+            totalBytes: effectiveTotalSize, uploadId: uploadId))
         return object
       } else if uploadResponse.statusCode == 308 {
         if let rangeHeader = uploadResponse.value(forHTTPHeaderField: "Range") {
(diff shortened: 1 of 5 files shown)


packages/storage/Tests/ResumableUploadTests.swift
```diff
@@ -761,38 +761,76 @@ import Testing
...
+    #expect(object.bucket == bucket)
+    #expect(object.size == Int64(data.count))
+  }
+
+  /// Tests resuming an interrupted upload for a fixed buffer in memory (`BytesSource`).
+  @Test func resumeUploadFixedBufferMemorySuccess() async throws {
+    let registry = MockRegistry.create()
+    let bucket = "test-bucket"
+    let objectName = "fixed-buffer-resumed"
+    let data = Data(repeating: 0xCD, count: 10 * 1024 * 1024)  // 10MB
```
coryan:
```suggestion
    let data = Data(repeating: 0xCD, count: 10 * 1024 * 1024)  // 10MiB
```

packages/storage/Tests/ResumableUploadTests.swift
```diff
@@ -761,38 +761,76 @@ import Testing
...
+  // --- Source Type 1: Fixed Buffers in Memory ---
+
+  /// Tests resumable upload using a fixed buffer in memory (`BytesSource`).
+  @Test func resumableUploadFixedBufferMemorySuccess() async throws {
     let registry = MockRegistry.create()
     let bucket = "test-bucket"
-    let objectName = "test-object"
-    let data = Data(repeating: 1, count: 1 * 1024 * 1024)  // 1MB payload
+    let objectName = "fixed-buffer-object"
+    let data = Data(repeating: 0xAB, count: 10 * 1024 * 1024)  // 10MB
```
coryan: nit:

```suggestion
    let data = Data(repeating: 0xAB, count: 10 * 1024 * 1024)  // 10MiB
```

packages/storage/Tests/ResumableUploadTests.swift
```diff
@@ -761,38 +761,76 @@ import Testing
...
+
+    // GCS reports 5MB uploaded (bytes 0-5242879 received)
+    let offset: Int64 = 5 * 1024 * 1024
+    let lastByte = offset - 1
 
     registry.register(
       response: .success(
         statusCode: 308, data: Data(),
-        headers: ["Range": "bytes=0-4999"]),
+        headers: ["Range": "bytes=0-\(lastByte)"]),
```
coryan, line 829: Please add some test cases where the query returns "no bytes are committed"...

packages/storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
```diff
@@ -470,9 +489,13 @@ extension HTTPClient {
...
     request.applyCustomerSuppliedEncryptionHeaders(options.customerEncryptionKey)
 
-    let end = offset + Int64(data.count) - 1
     let totalStr = totalSize.map { String($0) } ?? "*"
-    request.setValue("bytes \(offset)-\(end)/\(totalStr)", forHTTPHeaderField: "Content-Range")
+    if data.isEmpty {
+      request.setValue("bytes */\(totalStr)", forHTTPHeaderField: "Content-Range")
+    } else {
+      let end = offset + Int64(data.count) - 1
+      request.setValue("bytes \(offset)-\(end)/\(totalStr)", forHTTPHeaderField: "Content-Range")
```
coryan, line 496: Add a test case where the data size is unknown, but turns out to be 0 bytes (an empty stream). How would you format this header in that case?

Also, consider a test case where the last chunk is empty.

coryan (changes requested):
I think you have a bug for the empty stream case. Also streams that are exactly a multiple of the chunk size can be troublesome.

packages/storage/Tests/ResumableUploadTests.swift
chingor13: We have a `resumeZeroBytesOnServer()` test. In this case, the query response does not return a `Range` header

packages/storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
chingor13: Added tests for dynamic content where both last chunk is empty and a completely empty stream. Also verified with an integration test that succeeds for both these scenarios.

packages/storage/Tests/ResumableUploadTests.swift
coryan: Ack, thanks.

coryan (approved):

PR #120, opened 2026-08-04
cleanup(auth): use `async-http-client` package

### LINKED ISSUE #81: Switch `GoogleCloudAuth` to use `AsyncHTTPClient`
We are reliably informed that https://github.com/swift-server/async-http-client is the right thing to use. We use `URLSession` in a number of places, including

- [x] auth: move `AuthHttpClient` to `AsyncHTTPClient`
- [x] auth: move the tests to use `AsyncHTTPClient`

coryan:
Switch the (internal only) AuthHTTPClient type to use `AsyncHTTPClient.HTTPClient`.

Most of the changes are in the tests: the new implementation more easily supports mocks, no need for global state.

I refactored the mock to a single file, and use a `Deque` of closures to control the behavior of the mock. They can be async, throw errors, if you use named closures you can also repeat the same action a few times.

Fixes #81

### DIFF
--- a/packages/auth/Sources/GoogleCloudAuth/Http/AuthHTTPClient.swift
+++ b/packages/auth/Sources/GoogleCloudAuth/Http/AuthHTTPClient.swift
@@ -14,6 +14,7 @@
 
 import Foundation
 import AsyncHTTPClient
+import NIOFoundationCompat
 import struct Logging.Logger
 import struct NIOCore.TimeAmount
 import struct NIOCore.ByteBufferAllocator
@@ -98,7 +99,7 @@ struct AuthHTTPClient: Sendable {
       try self.ensureSuccess(response)
 
       do {
-        return try self.makeDecoder().decode(T.self, from: buffer.getData(at: 0, length: buffer.readableBytes)!)
+        return try self.makeDecoder().decode(T.self, from: Data(buffer: buffer))
       } catch let error as DecodingError {
         throw AuthHTTPError.decodingError(error: error)
       }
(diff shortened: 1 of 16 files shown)


packages/auth/Sources/GoogleCloudAuth/Http/AuthHTTPClient.swift
```diff
@@ -13,32 +13,70 @@
...
+import struct Logging.Logger
+import struct NIOCore.TimeAmount
+import struct NIOCore.ByteBufferAllocator
+
+protocol HTTPClientProtocol: Sendable {
+  func execute(
+    request: HTTPClientRequest,
+    timeout: Duration,
+    logger: Logger?
+  ) async throws -> HTTPClientResponse
```
chingor13, line 27: Do we want to have our own return type so we don't return a type from another library ("third party" but maybe core)? Essentially, putting this type in the interface locks us to this implementation.

coryan: In this particular case, I am not going to worry: this is an internal-only type to the GoogleCloudAuth module.

We will need to do something similar for GoogleCloudGax. There, I am not sure. On the one hand, wrapping all the types would seem to provide more insulation from changes. OTOH, avoiding all type leaks is a heavy lift (consider HTTP versions and status codes), and the types would still be `@_npi(GoogleCloudInternal)`.

I am also starting to think about configuration for these HTTPClient things. There are a lot of [configuration parameters](https://swiftpackageindex.com/swift-server/async-http-client/1.36.0/documentation/asynchttpclient/httpclient/configuration) for HTTPClient, we definitely do not want to expose those in the public API, but we do want to expose some of that functionality.

chingor13 (approved):

Issue #17, opened 2026-07-25
Storage: download APIs


coryan:
We want to write a storage client.  One of the main functions is to download objects. This client must:

- [x] It should be possible to specify the bucket name, object name, and (optional) generation
- [x] When possible (full reads, no decompression) the client should automatically perform CRC32C checksums
- [x] Optionally applications can request automatic MD5 checksums
- [x] Downloads should resume if interrupted, they resume from the last received bytes

This is similar to the Rust client APIs and may be translatable.

chingor13:
All subissues completed

PR #149, opened 2026-08-04
design(storage/download): add initial download API design

chingor13:
Towards #17

### DIFF
--- a/doc/design/storage_download_api_design.md
+++ b/doc/design/storage_download_api_design.md
@@ -1,0 +1,275 @@
+# Design Document: GCS Swift Client Library Download API
+
+This document outlines the API design for downloading objects using the Google Cloud Storage (GCS) Swift client library.
+
+## Goal
+
+Provide a language-idiomatic, memory-safe, high-performance, and resilient API for reading and downloading object contents from Cloud Storage in Swift.
+
+---
+
+## 1. Client Surface & Async Streaming Interface
+
+We want to expose a simple, intuitive client interface for reading object contents while leveraging Swift Concurrency's `AsyncSequence` for streaming data.
+
+### Decision
+The primary client method for downloading object content is `readObject`:
+
+```swift
+func readObject(
+  from bucket: String,
+  object: String,
+  options: ReadObjectOptions
+) -> ReadObjectSequence
+```
+
+- **Return Type:** Returns a custom struct `ReadObjectSequence` that conforms to `AsyncSequence` where `Element == Data`.
+- **Non-blocking Initialization:** Calling `readObject` immediately returns the `ReadObjectSequence` handle without blocking. Network requests and data streaming begin when the caller iterates over the sequence using `for try await chunk in client.readObject(...)`.
+
+---
+
+## 2. Target Object & Read Options
+
+Developers need full control over target object selection, precondition checks, and encryption options.
+
+### Key Options in `ReadObjectOptions`
+1. **Target Object Identification:**
+   - `bucket`: Bucket name containing the object.
+   - `object`: Object name/path within the bucket.
+   - `generation`: Optional object generation (`Int64?`) to read a specific revision of an object.
+2. **Preconditions:**
+   - Leverages `StoragePreconditions` (`ifGenerationMatch`, `ifGenerationNotMatch`, `ifMetagenerationMatch`, `ifMetagenerationNotMatch`) to ensure operations execute only when condition constraints pass.
+3. **Customer-Supplied Encryption Keys (CSEK):**
+   - Leverages `CustomerEncryptionKeyOptions` to send required encryption headers (`x-goog-encryption-algorithm`, `x-goog-encryption-key`, `x-goog-encryption-key-sha256`) when reading customer-encrypted objects.
+
+---
+
+## 3. Ranged Reads & Byte Ranges
+
+To support partial object downloads, parallel chunk downloads, or reading file footers (e.g. Parquet metadata), the API must support flexible ranged reads.
+
+### The `ReadRange` Abstraction
+
+Ranged reads are configured via the `ReadRange` enum:
+
+```swift
+public enum ReadRange: Sendable, Hashable, Equatable {
+  /// Read the entire object (default).
+  case entire
+
+  /// Read all bytes starting from `offset` to the end of the object (HTTP `bytes=N-`).
+  case fromOffset(Int64)
+
+  /// Read the last `count` bytes of the object (HTTP `bytes=-N`).
+  case suffix(Int64)
+
+  /// Read a bounded range of bytes from `start` to `end` inclusive (HTTP `bytes=start-end`).
+  case bounded(start: Int64, end: Int64)
+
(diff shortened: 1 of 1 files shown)


doc/design/storage_download_api_design.md
```diff
@@ -0,0 +1,275 @@
...
+```swift
+public enum ReadRange: Sendable, Hashable, Equatable {
+  /// Read the entire object (default).
+  case entire
+
+  /// Read all bytes starting from `offset` to the end of the object (HTTP `bytes=N-`).
+  case fromOffset(Int64)
+
+  /// Read the last `count` bytes of the object (HTTP `bytes=-N`).
+  case suffix(Int64)
```
coryan: Consider:

```suggestion
  /// Read all bytes starting from `offset` to the end of the object (HTTP `bytes=N-`).
  case fromOffset(UInt64)

  /// Read the first `count` bytes of the object (HTTP `bytes=0-N`).
  case prefix(UInt64)
  /// Read the last `count` bytes of the object (HTTP `bytes=-N`).
  case suffix(UInt64)
```

doc/design/storage_download_api_design.md
```diff
@@ -0,0 +1,275 @@
...
+
+  /// Read a bounded range of bytes from `start` to `end` inclusive (HTTP `bytes=start-end`).
+  case bounded(start: Int64, end: Int64)
```
coryan: Prefer using `UInt64` here, even if the underlying protocol uses `Int64`. We want the surface to be hard to use wrong.

doc/design/storage_download_api_design.md
```diff
@@ -0,0 +1,275 @@
...
+---
+
+## 5. Checksum Validation (CRC32C & MD5)
+
+To guarantee data integrity during transfer, Cloud Storage provides CRC32C and MD5 hashes in response headers (`x-goog-hash` or `ETag`).
+
+### Rules & Defaults
+1. **CRC32C (Default Enabled):** When performing full object reads (without ranged bounds or decompressive transcoding), the client automatically accumulates CRC32C checksums of received chunks on-the-fly and compares against `x-goog-hash` upon stream completion.
+2. **MD5 (Optional):** Applications can explicitly enable MD5 validation in `DownloadChecksumOptions`.
+3. **Automatic Bypass:** Checksum validation is automatically skipped when performing partial/ranged reads or when decompressive transcoding is active, because GCS header hashes reflect the entire raw object payload.
```
coryan, line 156: I think there may be some new thing to allow checksums on ranges, please double check.

doc/design/storage_download_api_design.md
```diff
@@ -0,0 +1,275 @@
...
+
+```swift
+func readObject(
+  from bucket: String,
+  object: String,
+  options: ReadObjectOptions
+) -> ReadObjectSequence
+```
+
+- **Return Type:** Returns a custom struct `ReadObjectSequence` that conforms to `AsyncSequence` where `Element == Data`.
```
coryan: This type should also return key data about the downloaded object, such as the generation picked (if the application does not provide one, it is good to find out which generation is returned). There are a number of other object attributes that are included as headers... but not the full contents of the `Object` resources. We decided to call these the `ObjectHighlights` in Rust.

doc/design/storage_download_api_design.md
```diff
@@ -0,0 +1,275 @@
...
+---
+
+## 6. Resumability & Error Handling
+
+Transient network failures during large object downloads should not require restarting the entire transfer from byte 0.
+
+### Resumption Protocol Strategy
+1. **Offset Tracking:** As chunks of `Data` are yielded by the `AsyncIterator`, the iterator tracks total `bytesReceived`.
+2. **Re-connection Range:** On a transient connection drop or socket error, if `autoResume` is enabled (default `true`), the iterator transparently initiates a new HTTP GET request requesting range `bytes={rangeStart + bytesReceived}-`.
+3. **Generation Lock:** To avoid reading corrupted or mismatched data if the object was updated mid-download, the resumption request includes `ifGenerationMatch={generation}` using the object's generation returned in the initial response header.
```
coryan: No, the resumption should use `generation=...` explicitly requesting the generation fetched during the download.  On buckets with multiple versions you could continue to download generation 1 even if the current generation is now 7. `ifGenerationMatch=1` would fail, but `generation=1` would succeed.

doc/design/storage_download_api_design.md
```diff
@@ -0,0 +1,275 @@
...
+
+### Options & Range Types
+
+```swift
+/// Specifies a byte range for ranged reads.
+public enum ReadRange: Sendable, Hashable, Equatable {
+  case entire
+  case fromOffset(Int64)
+  case suffix(Int64)
+  case bounded(start: Int64, end: Int64)
```
coryan: See my comments about `UInt64` vs. `Int64`...

doc/design/storage_download_api_design.md
```diff
@@ -0,0 +1,275 @@
...
+
+### `DownloadChecksumOptions` Interface
+
+```swift
+public struct DownloadChecksumOptions: Sendable, Hashable, Equatable {
+  /// Automatically validate CRC32C checksum on full reads.
+  public var crc32c: Bool
+
+  /// Automatically validate MD5 checksum when requested.
+  public var md5: Bool
```
coryan: I think in C++ we support "check against this known value of CRC32C", but I am not sure if it is that useful.  I mean, you get that from the object highlights and can compare in the application? 

Give it a thought.

coryan (changes requested):

doc/design/storage_download_api_design.md
chingor13: So chunk-level checksums are available if using the gRPC `ReadObject` RPC https://docs.cloud.google.com/storage/docs/data-validation#reads

Not all of our other client libraries support gRPC for storage. C++, go, and java are opt-in while rust uses both (gRPC for bidi streaming, http/rest for standard reads/writes)

This seems like a feature we can try to add later and not in the initial read_object implementation.

doc/design/storage_download_api_design.md
chingor13: So the metadata is returned after the initial data request is sent. This means that we will want to have tthe `readObject` method be async and we cannot defer network requests until trying to read the bytes.

chingor13: This should be fine because a user would probably never want to initialize the readObject request without wanting any network requests at all.

doc/design/storage_download_api_design.md
chingor13: Yes, this makes sense. We will continue downloading the version that was initially downloaded. We will only set the additional generation=X if the initial request did not include that filter.

doc/design/storage_download_api_design.md
chingor13: Updated this to re-use the `ChecksumOptions` type that we use for uploads. This lets a user specify `.auto` validation or with a known string value.

coryan (approved):

PR #165, opened 2026-08-05
feat(wkt): GoogleCloudWkt.Any conversion support

nolanleastin:


packages/wkt/Sources/GoogleCloudWktConvert/Any+Convert.swift
```diff
@@ -0,0 +1,29 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+import GoogleCloudWkt
+import SwiftProtobuf
+
+extension GoogleCloudWkt.`Any` {
+  public init(proto: SwiftProtobuf.Google_Protobuf_Any) throws {
+    let json = try proto.jsonUTF8Data()
```
coryan, line 21: LGTM.  Note: I suspect this `try` is load-bearing. In general, the contents of a `Google_Protobuf_Any` may correspond to an object type that is not even know to the local library. Extracting  JSON from of the object requires having the object descriptor....

Ditto on the `toProto()` function. Converting to the binary representation requires having the object descriptor.

coryan (approved):

packages/wkt/Sources/GoogleCloudWktConvert/Any+Convert.swift
nolanleastin: 👍

PR #170, opened 2026-08-05
feat(recaptchaenterprise/v1): generate library

### LINKED ISSUE #207: generate library for `google/cloud/recaptchaenterprise/v1`

AlexGhiondea:
Fixes #207


librarian.yaml
```diff
@@ -550,6 +550,11 @@ libraries:
     copyright_year: "2026"
     swift:
       library_name_override: GoogleCloudParameterManagerV1
+  - name: google-cloud-recaptchaenterprise-v1
+    version: 0.0.0-preview
+    copyright_year: "2026"
+    swift:
+      library_name_override: GoogleCloudRecaptchaEnterpriseV1
```
chingor13, line 557: The name of this service is "reCAPTCHA Enterprise" -- is "RecaptchaEnterprise" what we want?

chingor13 (approved):
non-blocking, but maybe worth a discussion on the name for this weirdly cased service?

librarian.yaml
AlexGhiondea: Good point. I think the other languages do use the weirdly cased `reCAPTCHA`. I'll update it.

AlexGhiondea: Actually, I stand corrected... The other languages do show the `reCAPTCHA` in the readme but not in the code... We can keep it as-is in the code and update the readme?

chingor13: Yeah, this is probably fine. Ruby and .NET use "RecaptchaEnterprise" in the library name/namespace

coryan: > Actually, I stand corrected... The other languages do show the `reCAPTCHA` in the readme but not in the code... We can keep it as-is in the code and update the readme?

Let's leave the README as-is...

Swift has different naming conventions for acronyms vs. other languages.  As an example, most languages would style `HTTP` as `Http`, Swift as `HTTP`. 

`CAPTCHA` is an acronym. So the Swift style would be `ReCAPTCHAEnterprise`.... IMHO less readable than `ReCaptchaEnterprise`, but that is what they do.

We can make any decision we want here, the easiest (in my mind) is to follow Swift's guidelines (`ReCAPTCHAEnterprise`). Weird, but consistent with the other libraries in the Swift SDK and with the Swift ecosystem.

We can still change our minds until GA if that makes things easier. Maybe open a "review weird names" bug to track these things and merge with the ALLCAPS name?

AlexGhiondea: I created this https://github.com/googleapis/google-cloud-swift/issues/192. 

So let's leave this the way it was generated for now and then do a pass where we cover all the other cases as well.

coryan: SG. :shipit:

PR #175, opened 2026-08-05
feat(storage): use configured retry policy and backoff policy for upload loop

### LINKED ISSUE #166: [storage upload] should use configured retry and backoff policies

chingor13:
Fixes https://github.com/googleapis/google-cloud-swift/issues/166

We wrap the entire resumable flow with the retry loop and keep track of status. For example, if we need to start an upload session, that is part of the retry loop. If we need to query partial upload status, that is part of the retry loop as well as the inner loop of uploading chunks of data.

We have 2 different paths -- if we have a seekable source or not.
* If seekable, if we are retrying a chunk that failed, we can seek to as far as the data is needed
* If non-seekable, we resend the current bytes we have buffered which may already be on the server (but will be ignored). This case is not as optimized but should still work.

### DIFF
--- a/packages/storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
+++ b/packages/storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
@@ -21,6 +21,12 @@ import GoogleCloudWkt
 @_spi(GoogleCloudInternal) import struct GoogleCloudGax._CRC32C
 import Crypto
 
+package enum ResumableUploadStatus: Sendable {
+  case unknown
+  case inprogress(UInt64)
+  case done(StorageObject)
+}
+
 extension StorageClient {
   /// Core upload method accepting any upload source.
   ///
@@ -37,6 +43,17 @@ extension StorageClient {
     options: UploadOptions = .default
   ) -> UploadTask {
     let clientOptions = self.options.client
+    let effectiveRetryPolicy =
+      options.retryPolicy ?? self.options.upload.retryPolicy
+      ?? clientOptions.retryPolicy
+    let effectiveBackoffPolicy =
+      options.backoffPolicy ?? self.options.upload.backoffPolicy ?? clientOptions.backoffPolicy
+    let retryLoop = _RetryLoop(
+      retryPolicy: effectiveRetryPolicy,
+      backoffPolicy: effectiveBackoffPolicy,
+      retryThrottler: clientOptions.retryThrottler,
+      idempotent: true
+    )
     return UploadTask.create { continuation in
       let httpClient = try HTTPClient(
         from: clientOptions, withDefaultEndpoint: StorageClient.defaultEndpoint)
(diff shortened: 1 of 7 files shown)


packages/storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
```diff
@@ -103,6 +111,49 @@ extension StorageClient {
     return object
   }
 
+  // Start a resumable upload session with configured retry and backoff. Upon success,
+  // returns the location of the upload (the uploadId)
+  fileprivate static func startResumableUpload(
```
coryan: Note to self: this is exactly what we want. Creating the session requires a retry loop, and that loop should be configured by the retry + backoff policies.

packages/storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
```diff
@@ -150,61 +203,110 @@ extension StorageClient {
...
         (isLast && totalSize == nil) ? (offset + Int64(chunk.count)) : totalSize
 
-      let uploadRequest = try await httpClient.buildUploadChunkRequest(
-        uploadId: uploadId, data: chunk, offset: offset, totalSize: effectiveTotalSize,
-        options: options,
-        checksum: checksum)
-      let (uploadData, uploadResponse) = try await httpClient.data(for: uploadRequest)
+      let res: ChunkUploadResult
+      do {
+        res = try await retryLoop.run { _ in
```
coryan: I don't think this is what we want.... This resets the retry attempt counter after each error ... Consider a retry policy that says "try 3 times" and the following sequence of errors:

- POST -> 200
- PUT (DATA) -> 429
- PUT (QUERY) -> 200
- PUT(DATA) -> 429
- PUT (QUERY) -> 200
- PUT (DATA) -> 429
- PUT (QUERY) -> 200
- ...

I do not believe this stops after the first error?


I think we want something like this (peudo codish):

```swift
var uploadStatus: UploadStatus = .inprogress(0) 
res = try await retryLoop.run { _ in
  if uploadStatus == .unknown {
    // If this fais
    uploadStatus = try await queryUploadStatus(...)
  }
  swift uploadStatus {
  case .unknown:
    throw crazyTheQueryUploadStatusFunctionNeverReturnsThis(...)
  case  .done(let object):
    return object
  case .inprogress(let committedBytes):
    // Adjust the current chunk and/or stream to avoid resending the same bytes
    uploadStatus = .unknown // if the thing throws that is the next status
    uploadStatus = try await sendNextChunk(...)
  }
}
```

where:

```swift
// A package-internal type
enum UploadStatus {
  case .unknown // lost track of the last committed byte
  case .inprogress(UInt64) // in progress with known number of bytes committed
  case .done(Object)
}
```

I am not sure that is the right structure, but I think that is the right idea, namely that the retry policy applies to the full upload, and errors in query or chunks count against the error budget.

coryan (changes requested):

packages/storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
chingor13: Ok, this feature is now refactored:

We wrap the entire resumable flow with the retry loop and keep track of status. For example, if we need to start an upload session, that is part of the retry loop. If we need to query partial upload status, that is part of the retry loop as well as the inner loop of uploading chunks of data.

We have 2 different paths if we have a seekable source or not. If seekable, if we are retrying a chunk that failed, we can seek to as far as the data is needed. If non-seekable, we resend the current bytes we have buffered which may already be on the server (but will be ignored). This case is not as optimized but should still work.

packages/storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
```diff
@@ -95,185 +178,374 @@ extension StorageClient {
...
-    return object
+
+    return try await retryLoop.run { _ in
+      let (responseData, response): (Data, HTTPURLResponse)
+      do {
+        (responseData, response) = try await httpClient.data(for: request)
+      } catch {
+        throw RequestError.io(error)
+      }
+      if response.statusCode == 503 {
```
coryan, line 189: 503 is a retryable error in storage-land (weird, I know):

https://docs.cloud.google.com/storage/docs/json_api/v1/status-codes#503_Service_Unavailable

In C++ and Rust we created a different default retry policy for storage to handle this weirdness. Maybe something to fix in a future PR?

packages/storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
```diff
@@ -95,185 +178,374 @@ extension StorageClient {
...
+    let (startData, startResponse): (Data, HTTPURLResponse)
+    do {
+      (startData, startResponse) = try await httpClient.data(for: startRequest)
+    } catch {
+      throw RequestError.io(error)
+    }
     guard startResponse.statusCode == 200,
       let location = startResponse.value(forHTTPHeaderField: "Location")
     else {
+      if startResponse.statusCode == 503 {
```
coryan, line 218: Ditto, this should be a retryable error.

coryan (approved):

packages/storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
chingor13: Tracking in #306

packages/storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
chingor13: Tracking in #306

PR #178, opened 2026-08-05
feat(maintenance/api/v1): generate library

### LINKED ISSUE #203: generate library for `google/cloud/maintenance/api/v1`

AlexGhiondea:
Fixes #203


librarian.yaml
```diff
@@ -519,6 +519,9 @@ libraries:
   - name: google-cloud-lustre-v1
     version: 0.0.0-preview
     copyright_year: "2026"
+  - name: google-cloud-maintenance-api-v1
```
coryan, line 522: Should we override to `GoogleCloudMaintenanceAPIV1` ?

I think we used `Api` in other places, so maybe just add this to the "revisit later" bug?

coryan (approved):

PR #189, opened 2026-08-05
feat(storage): add readObject options and stubs

### LINKED ISSUE #191: [storage download] Create simple client-surface to "read the object contents"

chingor13:
Fixes #191

### DIFF
--- a/packages/storage/Sources/GoogleCloudStorage/DownloadOptions.swift
+++ b/packages/storage/Sources/GoogleCloudStorage/DownloadOptions.swift
@@ -1,0 +1,183 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+/// Specifies a byte range for ranged reads.
+public enum ReadRange: Sendable, Hashable, Equatable {
+  /// Read the entire object (default).
+  case entire
+
+  /// Read all bytes starting from `offset` to the end of the object (HTTP `bytes=N-`).
+  case fromOffset(UInt64)
+
+  /// Read the first `count` bytes of the object (HTTP `bytes=0-N`).
+  case prefix(UInt64)
+
+  /// Read the last `count` bytes of the object (HTTP `bytes=-N`).
+  case suffix(UInt64)
+
+  /// Read a bounded range of bytes from `start` to `end` inclusive (HTTP `bytes=start-end`).
+  case bounded(start: UInt64, end: UInt64)
+
+  /// Convenience initializer for Swift `ClosedRange<UInt64>`.
+  public init(_ range: ClosedRange<UInt64>) {
+    self = .bounded(start: range.lowerBound, end: range.upperBound)
+  }
+
+  /// Converts the range specification to an HTTP `Range` header value string.
+  public var headerValue: String? {
+    switch self {
+    case .entire:
+      return nil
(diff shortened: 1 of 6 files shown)


packages/storage/Sources/GoogleCloudStorage/DownloadOptions.swift
```diff
@@ -0,0 +1,183 @@
...
+  case fromOffset(UInt64)
+
+  /// Read the first `count` bytes of the object (HTTP `bytes=0-N`).
+  case prefix(UInt64)
+
+  /// Read the last `count` bytes of the object (HTTP `bytes=-N`).
+  case suffix(UInt64)
+
+  /// Read a bounded range of bytes from `start` to `end` inclusive (HTTP `bytes=start-end`).
+  case bounded(start: UInt64, end: UInt64)
```
coryan, line 32: What happens if `end <= start` ?

coryan: An alternative to consider:  make this enum private to the package, and then only allow constructors that leave it in a consistent state.

That would be the Rusty way, type safe, but slightly more code and (maybe) harder to use.  I am not yet sure what is the Swifty way.  Thoughts?

packages/storage/Sources/GoogleCloudStorage/DownloadOptions.swift
```diff
@@ -0,0 +1,183 @@
...
+  public var preconditions: StoragePreconditions?
+
+  /// Options for Customer-Supplied Encryption Keys (CSEK).
+  public var customerEncryptionKey: CustomerEncryptionKeyOptions?
+
+  /// Byte range for partial/ranged reads. Defaults to `.entire`.
+  public var range: ReadRange = .entire
+
+  /// Flag to disable automatic decompressive transcoding by GCS. Defaults to `false`.
+  public var disableDecompressiveTranscoding: Bool = false
```
coryan: I think  `enableDecompressiveTranscoding: Bool = true` would be easier to grok.

packages/storage/Sources/GoogleCloudStorage/DownloadOptions.swift
```diff
@@ -0,0 +1,183 @@
...
+    generation: Int64 = 0,
+    metageneration: Int64? = nil,
+    etag: String? = nil,
+    crc32c: String? = nil,
+    md5Hash: String? = nil,
+    contentType: String? = nil,
+    contentEncoding: String? = nil,
+    contentDisposition: String? = nil,
+    storageClass: String? = nil,
+    updated: Date? = nil
```
coryan: Why not use the `with` trick? Because the fields are `let`?

packages/storage/Sources/GoogleCloudStorage/DownloadOptions.swift
```diff
@@ -0,0 +1,183 @@
...
+  public func with(_ config: (inout Self) -> Void) -> Self {
+    var copy = self
+    config(&copy)
+    return copy
+  }
+}
+
+/// Metadata attributes for an object returned in response headers during a download.
+public struct ReadObjectMetadata: Sendable, Hashable, Equatable {
+  public let bucket: String
```
coryan: We should document public fields. They go into the reference docs.

packages/storage/Sources/GoogleCloudStorage/DownloadOptions.swift
```diff
@@ -0,0 +1,183 @@
...
+    self.storageClass = storageClass
+    self.updated = updated
+  }
+}
+
+/// An asynchronous sequence of `Data` chunks representing an object payload being downloaded.
+public struct ReadObjectSequence: AsyncSequence, Sendable {
+  public typealias Element = Data
+
+  public let bucket: String
```
coryan, line 150: Ditto regarding public fields.  Consider making some of these fields private too.

coryan (approved):
Please fix the documentation or create tracking bugs for it.

The `disable... = false` thing I should have caught in the design review, please fix before merging this.

packages/storage/Sources/GoogleCloudStorage/DownloadOptions.swift
chingor13: We will follow up with more exploration in #219

PR #213, opened 2026-08-06
refactor(storage): use generated Object type

### LINKED ISSUE #27: [storage upload] use generated object metadata type
For now, the upload functionality is returning a handwritten Storage object type. If we are generating this type, we should look to use that type.

```
There are more variations: you may want to calculate and validate both.
You may want to provide a pre-calculated value (say computed when the file was downloaded).
```

chingor13:
Fixes #27 

This was partially done in #150


packages/storage/Sources/GoogleCloudStorage/Object+Compatibility.swift
```diff
@@ -0,0 +1,33 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+import GoogleCloudWkt
+
+// These are convenience methods for setting metadata and timestamp
+// fields.
+extension Object {
+  public var customMetadata: [String: String]? {
```
coryan: This will need docs. My advice is to remove them. Once we release the library, we can never take them back.

packages/storage/Sources/GoogleCloudStorage/Object+Codable.swift
```diff
@@ -0,0 +1,218 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+import GoogleCloudWkt
+
+extension Object {
```
coryan: Why do we need this?  The generator should (and does) provide a default implementation.

packages/storage/Sources/GoogleCloudStorage/Object+Codable.swift
```diff
@@ -0,0 +1,218 @@
...
+    case contentLanguage
+    case deleteTime
+    case timeDeleted
+    case finalizeTime
+    case contentType
+    case createTime
+    case timeCreated
+    case componentCount
+    case checksums
+    case crc32c
```
coryan: Oh, I see, this supports the v1 format as well as the v2 format. Neat.

One thing to keep in mind: the generator provides explicit decode / encode functions under certain conditions (if there is a oneof or a map with non-string types).  I think that will break in this case, but we could add a tweak / option to the generator to handle that case.

Can you document the fields that have overrides? It may help us in the future as we need to inject new fields.

coryan (changes requested):
I would remove the convenience methods, maybe in a follow up PR. They seem useful now, but they will be redundant in the future and just cause trouble.

Please add some comments to the Encodable implementation.

packages/storage/Sources/GoogleCloudStorage/Object+Compatibility.swift
chingor13: Removed

packages/storage/Sources/GoogleCloudStorage/Object+Codable.swift
chingor13: Alternatively, we can maintain the v1 response type and provide a converter from the v1 type to v2 Object

coryan: Your call. We can always fix this if needed.

chingor13: Rather than messing with the Object v2 parsing, we will explicitly parse the v1 response and convert to a v2 type

coryan (approved):

PR #226, opened 2026-08-07
refactor: move checksum options into their own file

chingor13:
These types will be shared between `UploadOptions` and `DownloadOptions` so let's put them in their own file


packages/storage/Sources/GoogleCloudStorage/ChecksumOptions.swift
```diff
@@ -0,0 +1,84 @@
...
+
+    /// Creates a `ChecksumValue` from an integer literal containing a CRC32C checksum value.
+    public init(integerLiteral value: UInt64) {
+      self.init(UInt32(truncatingIfNeeded: value))
+    }
+  }
+
+  /// Creates a new `ChecksumOptions` configuration for validating data with Google Cloud Storage.
+  ///
+  /// You can configure automatic on-the-fly calculation (`.auto`), pre-computed values (e.g., `.value("...")` or a string literal `"..."`),
```
coryan: Consider:

```suggestion
  /// You can configure:
  /// - `.auto`: automatic on-the-fly calculation,
  /// - `.value("...")` or a string literal `"..."`: pre-computed values
  ///
  /// You can also enable both crc32c and md5  checksums simultaneously.
```

coryan (approved):
Good as-is, but maybe consider improving the doc comments a bit.

PR #228, opened 2026-08-07
feat: implement basic readObject logic

### LINKED ISSUE #227: [storage download] client should return  `AsyncSequence<Data>`
Implement the basic downloading in the form of an `AsyncSequence<Data>`

chingor13:
Fixes #227

### DIFF
--- a/packages/gax/Sources/GoogleCloudGax/HttpClient.swift
+++ b/packages/gax/Sources/GoogleCloudGax/HttpClient.swift
@@ -59,8 +59,10 @@ public struct HTTPClient {
 
   public func Request(path: String, query: [URLQueryItem]) async throws -> URLRequest {
     var components = self.baseURL
-    components.path = path
-    components.queryItems = query
+    if !query.isEmpty {
+      components.queryItems = query
+    }
+    components.percentEncodedPath = percentEncodedPath
     guard let url = components.url else {
       throw RequestError.binding("bad URL for path=\(path), baseURL=\(self.baseURL)")
     }
--- a/packages/gax/Sources/GoogleCloudGax/HttpClient.swift
+++ b/packages/gax/Sources/GoogleCloudGax/HttpClient.swift
@@ -59,12 +59,28 @@ public struct HTTPClient {
 
   public func Request(path: String, query: [URLQueryItem]) async throws -> URLRequest {
     var components = self.baseURL
+    components.path = path
+    components.queryItems = query
+    guard let url = components.url else {
+      throw RequestError.binding("bad URL for path=\(path), baseURL=\(self.baseURL)")
+    }
+    var request = URLRequest(url: url)
+    let headers = try await self.credentials.headers()
+    for (key, value) in headers {
+      request.setValue(value, forHTTPHeaderField: key)
+    }
+    return request
+  }
+
+  public func Request(percentEncodedPath: String, query: [URLQueryItem]) async throws -> URLRequest
+  {
+    var components = self.baseURL
     if !query.isEmpty {
       components.queryItems = query
     }
     components.percentEncodedPath = percentEncodedPath
     guard let url = components.url else {
-      throw RequestError.binding("bad URL for path=\(path), baseURL=\(self.baseURL)")
+      throw RequestError.binding("bad URL for path=\(percentEncodedPath), baseURL=\(self.baseURL)")
     }
     var request = URLRequest(url: url)
     let headers = try await self.credentials.headers()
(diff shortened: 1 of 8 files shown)


packages/gax/Sources/GoogleCloudGax/HttpClient.swift
```diff
@@ -72,6 +72,24 @@ public struct HTTPClient {
     return request
   }
 
+  public func Request(percentEncodedPath: String, query: [URLQueryItem]) async throws -> URLRequest
+  {
+    var components = self.baseURL
+    if !query.isEmpty {
+      components.queryItems = query
+    }
+    components.percentEncodedPath = percentEncodedPath
```
coryan, line 81: TIL

packages/storage/Sources/GoogleCloudStorage/StorageClient+Download.swift
```diff
@@ -27,6 +31,157 @@ extension StorageClient {
     object: String,
     options: ReadObjectOptions = .init()
   ) async throws -> ReadObjectResult {
-    fatalError("Unimplemented")
+    let clientOptions = self.options.client
+    let httpClient = try HTTPClient(
+      from: clientOptions, withDefaultEndpoint: StorageClient.defaultEndpoint)
```
coryan: Why do we need to create a new `HTTPClient` for each download? I would like to reuse the client and its connection pool (if it has one) if possible. Can we do that?

packages/storage/Sources/GoogleCloudStorage/StorageClient+Download.swift
```diff
@@ -27,6 +31,157 @@ extension StorageClient {
...
+    let request = try await httpClient.buildReadObjectRequest(
+      bucket: bucket, object: object, options: options)
+    let (data, response) = try await httpClient.data(for: request)
+
+    guard (200..<300).contains(response.statusCode) else {
```
coryan, line 38: Presumably we will add retry loop here? No need to do that in this PR, just asking about the plan.

packages/storage/Sources/GoogleCloudStorage/StorageClient+Download.swift
```diff
@@ -27,6 +31,157 @@ extension StorageClient {
...
+    var request = try await self.Request(
+      percentEncodedPath: "/storage/v1/b/\(encodedBucket)/o/\(encodedObject)", query: queryItems)
+    request.httpMethod = "GET"
+
+    request.applyCustomerSuppliedEncryptionHeaders(options.customerEncryptionKey)
+
+    return request
+  }
+
+  fileprivate func parseReadObjectMetadata(
```
coryan: I am not sure why this is part of the `HTTPClient` does it use `self` anywhere?

This seems like it belongs in the Storage extension, maybe as `private` or `fileprivate` functions.

packages/storage/Tests/DownloadTests.swift
```diff
@@ -0,0 +1,184 @@
...
+      lastReq?.value(forHTTPHeaderField: "x-goog-encryption-key-sha256") == csek.keyHashBase64)
+
+    var downloaded = Data()
+    for try await chunk in result.body {
+      downloaded.append(chunk)
+    }
+    #expect(downloaded == payload)
+  }
+
+  @Test func downloadObjectWithSlashesInPath() async throws {
```
coryan: Should we have more tests for other object names that require URL encoding? What about `?` and `&`?

coryan (changes requested):

packages/storage/Sources/GoogleCloudStorage/StorageClient+Download.swift
chingor13: Yes, we will layer the retries on later. This should be a lot easier than the retries around the upload path.

packages/storage/Sources/GoogleCloudStorage/StorageClient+Download.swift
chingor13: Fixed to use the one stored as .inner from init()

packages/storage/Sources/GoogleCloudStorage/StorageClient+Download.swift
chingor13: Moved to the StorageClient extension

coryan (approved):

packages/storage/Tests/DownloadTests.swift
chingor13: Added a few more unit and integration tests

PR #309, opened 2026-08-10
feat(storage): qualify external conversion extensions

nolanleastin:


packages/storage/Sources/GoogleCloudStorage/generated/StorageControl/Convert/GoogleRpc/Status+Convert.swift
```diff
@@ -22,7 +22,7 @@ import GoogleCloudWkt
 import GoogleRpc
 internal import GoogleCloudWktConvert
 
-extension Status {
+extension GoogleRpc.Status {
   internal typealias ProtoType = StorageControlProtos.Google_Rpc_Status
 
   internal init(proto: ProtoType) throws {
```
coryan, line 28: This is safe, because the `ProtoType = ...` is defined in this package, so it cannot conflict with a definition in a different package.

packages/storage/Sources/GoogleCloudStorage/generated/StorageControl/Convert/GoogleRpc/Status+Convert.swift
coryan, line 35: In contrast, if we had another package that provided an extension for `GoogleRpc.Status`, then (I believe, check my math) we would have a conflict because there are two definitions of `Google.Rpc.Status\toProto`, both internal to their packages, but both with the same "name".

I may be hyper sensitive to this stuff because Rust has a whole lot of machinery to prevent conflicting definitions of extensions (aka "the orphan rule").

No action needed, just something to think about for the future.

coryan (commented):
No action needed, just something for you and @chingor13 to think about.

packages/storage/Sources/GoogleCloudStorage/generated/StorageControl/Convert/GoogleRpc/Status+Convert.swift
nolanleastin: Yes you're 100% right. I should have been explicit about why I am adding this. 

There's a GoogleLongRunning.Operation which conflicts with Foundation.Operation. 

I was getting errors like these:

```
error: 'Operation' is ambiguous for type lookup in this context
note: candidate 1: Foundation.Operation
note: candidate 2: GoogleLongRunning.Operation
```

packages/storage/Sources/GoogleCloudStorage/generated/StorageControl/Convert/GoogleRpc/Status+Convert.swift
nolanleastin: right agree

coryan:
It may be best if we merge #305 first...

coryan (approved):

PR #325, opened 2026-08-11
ci: create minimum docs build

### LINKED ISSUE #302: ci: create a build for documents in generated libraries
AFAICT, the only way to build the documentation requires installing the [swift-docc-plugin](https://github.com/swiftlang/swift-docc-plugin) in your `Package.swift` file. Otherwise commands like `swift package generate-documentation` fail (they simply do not exist).

I don't want to install the plugin in all the package files, that seems like adding dependencies that the final consumers may not need.

So we need to be able to generate the docs from the top-level directory. That means all the packages must be targets at the top-level directory too.

We first need to add all the packages to the top-level `Package.swift` file.

coryan:
Validate the hand-crafted libraries and some generated libraries.
Validating all the generated libraries requires more machinery, and most
likely will succeed (or fail) in the same way.

Fixes #302 towards #12

### DIFF
--- a/ci/gcb/scripts/docs.sh
+++ b/ci/gcb/scripts/docs.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+#
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
+
+echo "--- SWIFT VERSION ---"
+swift --version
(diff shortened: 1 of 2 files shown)


coryan:
I am going to merge #326 first

ci/gcb/scripts/docs.sh
```diff
@@ -0,0 +1,72 @@
...
+
+# TODO(https://github.com/googleapis/google-cloud-swift/issues/308) - the xref links need fixing
+# after that we can merge this to the main loop
+echo; echo "================ Building GoogleCloudStorage ================"
+if swift package generate-documentation  --target "GoogleCloudStorage"; then
+    echo "✓ GoogleCloudStorage built successfully"
+else
+    echo; echo "✗ GoogleCloudStorage failed to build"
+    errors=$((errors + 1))
+    continue
```
nolanleastin: remove continue, this isn't a for loop

ci/gcb/scripts/docs.sh
```diff
@@ -0,0 +1,72 @@
...
+        echo; echo "✗ ${target} failed to build"
+        errors=$((errors + 1))
+        continue
+    fi
+done
+
+# TODO(https://github.com/googleapis/google-cloud-swift/issues/308) - the xref links need fixing
+# after that we can merge this to the main loop
+echo; echo "================ Building GoogleCloudStorage ================"
+if swift package generate-documentation  --target "GoogleCloudStorage"; then
```
nolanleastin, line 59: increment `count` here, too

nolanleastin (approved):

ci/gcb/scripts/docs.sh
coryan: Done.

ci/gcb/scripts/docs.sh
coryan: Fixed

PR #355, opened 2026-08-11
ci: support full builds

### LINKED ISSUE #352: ci: optionally enable all packages in the top-level `Package.swift`
We want to test document generation for any package. That requires either:

- Putting the docc plugin in each generated package or
- Making all the generated libraries available as targets of the top-level package.

We are going with the second option. The downside is that building with all the packages defined as dependencies is very slow, just building the dependency tree takes about 2 minutes. So we will make the list of packages optional, controlled by an environment variable.

By default only the normal dependencies and targets will be available. In the CI builds we will enable the environment variable, and then the top-level Package.swift file can dynamically search the `generated/*` directory and discover what packages are available.

coryan:
When the environment variable `GOOGLE_CLOUD_SWIFT_FULL_BUILD` is set to true, the top-level package discovers all the packages in `generated/*`. We can use this in CI builds to compile all the code in parallel, and to generate the documentation for any generated library.

Fixes #352

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -137,7 +137,10 @@ let package = Package(
 // A generated package description.
 //
 // This is just enough information to populate the `Package` data structure. It needs both the
-// package path and its
+// package path, its product [^1], and any (optional) traits that we enable for the tests in
+// `Tests/`.
+//
+// [^1]: in general, packages have many products, but our packages in `generated/*` only have one.
 struct Generated {
   public let name: String
   public let module: String
@@ -150,6 +153,7 @@ struct Generated {
   }
 }
 
+/// Finds the generated packages used for the build.
 func selectGeneratedPackages() -> [Generated] {
   let fullBuild = ProcessInfo.processInfo.environment["GOOGLE_CLOUD_SWIFT_FULL_BUILD"] == "true"
   if fullBuild {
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -118,3 +133,80 @@ let package = Package(
     ),
   ]
 )
+
+// A generated package description.
+//
+// This is just enough information to populate the `Package` data structure. It needs both the
+// package path and its
```
nolanleastin: and its ... what?!?! don't keep us in suspense

Package.swift
```diff
@@ -118,3 +133,80 @@ let package = Package(
...
+
+func selectGeneratedPackages() -> [Generated] {
+  let fullBuild = ProcessInfo.processInfo.environment["GOOGLE_CLOUD_SWIFT_FULL_BUILD"] == "true"
+  if fullBuild {
+    return generatedPackagesFull()
+  }
+  return generatedPackagesStatic()
+}
+
+func generatedPackagesStatic() -> [Generated] {
```
nolanleastin, line 170: nit: how did we pick these? maybe just add a short comment

nolanleastin (approved):

Package.swift
coryan: Done.

Package.swift
coryan: Done

Issue #219, opened 2026-08-06
[storage download] consider making range initializers more typesafe


chingor13:
https://github.com/googleapis/google-cloud-swift/pull/189#discussion_r3724750655

It might be possible to prevent invalid ranges using initializers rather than letting users use the configuration enums directly.

> An alternative to consider: make this enum private to the package, and then only allow constructors that leave it in a consistent state.
> That would be the Rusty way, type safe, but slightly more code and (maybe) harder to use. I am not yet sure what is the Swifty way. Thoughts?

chingor13:
We also need to make the request for the last 0 bytes an invalid construct -- the server will return everything.

PR #363, opened 2026-08-11
feat(storage): implement ranged reads

### LINKED ISSUE #354: [storage download] perform ranged reads (including "the last N bytes", and "all the bytes starting a N")

chingor13:
Fixes #354

Note: we are making the `Range` header (used in uploads) parsing code mirror the `Content-Range` header (used in downloads) parsing code.

### DIFF
--- a/packages/storage/Sources/GoogleCloudStorage/DownloadError.swift
+++ b/packages/storage/Sources/GoogleCloudStorage/DownloadError.swift
@@ -19,8 +19,8 @@ public enum DownloadError: Error, Sendable, Equatable {
   /// The downloaded payload checksum did not match the expected checksum.
   case checksumMismatch(expected: String, actual: String, algorithm: String)
 
-  /// The requested byte range specification is invalid or malformed.
-  case invalidRange(String)
+  /// The range header returned by GCS is invalid or malformed.
+  case invalidRangeHeader(String)
 
   /// Transparent download auto-resumption failed after a network interruption.
   case resumeFailed(bytesReceived: UInt64, message: String)
--- a/packages/storage/Sources/GoogleCloudStorage/DownloadError.swift
+++ b/packages/storage/Sources/GoogleCloudStorage/DownloadError.swift
@@ -19,7 +19,7 @@ public enum DownloadError: Error, Sendable, Equatable {
   /// The downloaded payload checksum did not match the expected checksum.
   case checksumMismatch(expected: String, actual: String, algorithm: String)
 
-  /// The range header returned by GCS is invalid or malformed.
+  /// The range header returned by Cloud Storage is invalid or malformed.
   case invalidRangeHeader(String)
 
   /// Transparent download auto-resumption failed after a network interruption.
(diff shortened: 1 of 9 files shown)


packages/storage/Sources/GoogleCloudStorage/DownloadError.swift
```diff
@@ -19,8 +19,8 @@ public enum DownloadError: Error, Sendable, Equatable {
   /// The downloaded payload checksum did not match the expected checksum.
   case checksumMismatch(expected: String, actual: String, algorithm: String)
 
-  /// The requested byte range specification is invalid or malformed.
-  case invalidRange(String)
+  /// The range header returned by GCS is invalid or malformed.
```
coryan:
```suggestion
  /// The range header returned by Cloud Storage is invalid or malformed.
```

packages/storage/Sources/GoogleCloudStorage/StorageClient+Download.swift
```diff
@@ -31,6 +31,21 @@ extension StorageClient {
...
+    if case .bounded(let start, let end) = options.range {
+      guard start <= end else {
+        throw DownloadError.invalidRangeHeader("Range start (\(start)) must be <= end (\(end)).")
+      }
+    }
+
+    // short-circuit if no data is needed
+    if case .prefix(0) = options.range {
+      return emptyReadResult(bucket: bucket, object: object, options: options)
+    }
```
coryan: Hmmm... an actual read may return an error, no? So this is not strictly a short-circuit, it returns different results that making an actual RPC.

A bit of a convoluted way to do things admittedly.

packages/storage/Tests/DownloadTests.swift
```diff
@@ -192,4 +192,239 @@ import Testing
...
+      options: ReadObjectOptions().with { $0.range = .prefix(0) }
+    )
+    #expect(prefixResult.metadata.size == 0)
+    var prefixData = Data()
+    for try await chunk in prefixResult.body {
+      prefixData.append(chunk)
+    }
+    #expect(prefixData.isEmpty)
+
+    // Suffix 0
```
coryan, line 436: I seem to recall some odd behavior with these ranges where "0" is interpreted as "return everything"... maybe worth checking the behavior of the service and at least document it.

coryan (approved):
A few nits and then the thing about short-circuits that seem to skip the RPC checking if the object exists at all.

Oh, reading at least one byte gives you the metadata generation, generation,  other things attributes, no?

packages/storage/Tests/DownloadTests.swift
chingor13: Range headers 0-0 actually means give me the first byte. There is no way to specify a ranged request of 0 bytes -- so we can request the minimum and just not yield it to the caller.

There is also weird behavior for suffix of 0 where the server will return everything. We should add this to the list of range validations as part of #219. (The c++ implementation rejects this type as invalid https://github.com/googleapis/google-cloud-cpp/blob/abc95d540b1faaa0ce6fb40b6b960d4df5ff1d5b/google/cloud/storage/internal/grpc/object_request_parser.cc#L420-L422)

PR #366, opened 2026-08-11
ci: specify directory for packages/ integration tests

### LINKED ISSUE #365: ci: should attempt to compile and run storage integration tests
Currently, storage integration tests do not compile

chingor13:
Fixes #365 

Note: Most of the storage tests are skipped unless you provide configuration environment variables. This PR ensures that we do compile the tests.

### DIFF
--- a/ci/gcb/scripts/integration-tests.sh
+++ b/ci/gcb/scripts/integration-tests.sh
@@ -47,15 +47,13 @@ fi
 for dir in packages/*; do
     [[ -f "${dir}/Package.swift" ]] || continue
     count=$((count + 1))
-    pushd "${dir}"
     echo "--- Running ${dir} integration tests ---"
-    if swift test "${flags[@]}" --quiet --traits IntegrationTests; then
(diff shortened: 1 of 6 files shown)


ci/gcb/scripts/integration-tests.sh
```diff
@@ -47,13 +47,15 @@ fi
 for dir in packages/*; do
     [[ -f "${dir}/Package.swift" ]] || continue
     count=$((count + 1))
+    pushd "${dir}"
     echo "--- Running ${dir} integration tests ---"
     if swift test "${flags[@]}" --quiet --traits IntegrationTests; then
```
coryan: Does this work? I think it is nicer to use the flag instead of messing around with the cwd, and using `--enable-all-traits` should work even if there are no traits.



```suggestion
    echo "--- Running ${dir} integration tests ---"
    if swift test "${flags[@]}" --quiet --package-path "${dir}" --enable-all-traits; then
```

coryan (changes requested):

coryan (approved):

PR #405, opened 2026-08-12
feat(devtools-containeranalysis-v1): generate library

### LINKED ISSUE #397: generate library for `google/devtools/containeranalysis/v1`

coryan:
Fixes #397


chingor13 (approved):

chingor13 (changes requested):
Actually, I remember this one being special with some kind of compatibility layer with Grafeas. Do we care about that?

generated/google-devtools-containeranalysis-v1/Package.swift
```diff
@@ -0,0 +1,47 @@
...
+  name: "GoogleCloudContainerAnalysisV1",
+  platforms: [.macOS(.v15)],
+  products: [
+    .library(name: "GoogleCloudContainerAnalysisV1", targets: ["GoogleCloudContainerAnalysisV1"])
+  ],
+  dependencies: [
+    .package(path: "../../packages/auth"),
+    .package(path: "../../packages/gax"),
+    .package(path: "../../packages/wkt"),
+    .package(path: "../../generated/grafeas-v1"),
```
coryan, line 30: It does depend on grafeas... `grafeas-v1` is already configured in the well-known dependencies, so sidekick picks it up automatically.

chingor13: Ok, I remember us doing something weird with synthtool replacements to make it possible to pass a `ContainerAnalysisClient` as a `GrafeasClient`, but it's probably unnecessary

chingor13 (approved):

generated/google-devtools-containeranalysis-v1/Package.swift
coryan: Oh, never supported that in C++ or Rust.  I think we should stick to plain GAPIC for all these things.

PR #407, opened 2026-08-12
feat(dns-v1): generate library

### LINKED ISSUE #362: generate library for `google/cloud/dns/v1`

coryan:
This is a discovery-based library, so the changes to `librarian.yaml` are hand-crafted. It also has some discovery-based LROs, which require hand-crafting the `Operation+Additions.swift` file.

Fixes #362


librarian.yaml
```diff
@@ -568,6 +568,25 @@ libraries:
       default_traits:
         - SearchService
         - RecommendationService
+  - name: google-cloud-dns-v1
+    version: 0.0.0-preview
+    apis:
+      - path: discoveries/dns.v1.json
+    copyright_year: "2026"
+    keep:
+      - Sources/GoogleCloudDNSV1/Operation+Additions.swift
```
chingor13, line 577: I don't see this file yet? Is this coming later?

chingor13 (changes requested):

librarian.yaml
coryan: I see it:

image

chingor13: Ok, I couldn't find it on the review page because of the number of files and the way the UI loads the diff

chingor13 (approved):

PR #410, opened 2026-08-12
test(storage): use shared test fixture for integration tests

chingor13:
Some ranged read tests need to create a file to read. We test several ranges over the same data so we shouldn't need write the same file multiple times.

This uses a shared class-level setup for creating the fixture once. We use `Task` because we cannot run async create options as a class-level variable.

### DIFF
--- a/packages/storage/Tests/IntegrationTests/StorageClientIntegrationTests.swift
+++ b/packages/storage/Tests/IntegrationTests/StorageClientIntegrationTests.swift
@@ -18,7 +18,8 @@ import Testing
 
 #if IntegrationTests
 
-  @Suite(.enabled(if: ProcessInfo.processInfo.environment["GOOGLE_CLOUD_PROJECT"] != nil))
+  @Suite(.enabled(if: ProcessInfo.processInfo.environment["GOOGLE_CLOUD_PROJECT"] != nil &&
+  ProcessInfo.processInfo.environment["GOOGLE_CLOUD_SWIFT_TEST_BUCKET"] != nil` ?))
   struct StorageClientIntegrationTests {
     @Test func testFileUpload() async throws {
       let bucketName =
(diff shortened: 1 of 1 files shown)


packages/storage/Tests/IntegrationTests/StorageClientIntegrationTests.swift
```diff
@@ -18,12 +18,9 @@ import Testing
 
 #if IntegrationTests
 
-  @Suite struct StorageClientIntegrationTests {
+  @Suite(.enabled(if: ProcessInfo.processInfo.environment["GOOGLE_CLOUD_PROJECT"] != nil))
```
coryan:
```suggestion
  @Suite(.enabled(if: ProcessInfo.processInfo.environment["GOOGLE_CLOUD_PROJECT"] != nil &&
  ProcessInfo.processInfo.environment["GOOGLE_CLOUD_SWIFT_TEST_BUCKET"] != nil` ?))
```

And the remove the `test-bucket` thing inside because bucket names are global

packages/storage/Tests/IntegrationTests/StorageClientIntegrationTests.swift
```diff
@@ -813,6 +653,114 @@ import Testing
...
+    @Test(arguments: [
+      (ReadObjectRange.bounded(start: 10, end: 19), "abcdefghij"),
+      (ReadObjectRange.fromOffset(36), "ABCDEFGHIJKLMNOPQRSTUVWXYZ"),
+      (ReadObjectRange.prefix(10), "0123456789"),
+      (ReadObjectRange.suffix(10), "QRSTUVWXYZ"),
+      (ReadObjectRange(5...15), "56789abcdef"),
+      (ReadObjectRange.entire, "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"),
+    ])
+    func testRangedDownload(range: ReadObjectRange, expectedContent: String) async throws {
+      let fixture = try await Self.sharedFixture.value
```
coryan, line 669: TIL

coryan (approved):

PR #422, opened 2026-08-13
refactor(storage): use gax's _HttpClient

### LINKED ISSUE #421: [storage] switch to use _HttpClient

chingor13:
Fixes #421


packages/storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
```diff
@@ -178,16 +188,16 @@ extension StorageClient {
...
+      let response: _HTTPClientResponse
       do {
-        (responseData, response) = try await httpClient.data(for: request)
+        response = try await request.execute()
       } catch {
         throw RequestError.io(error)
       }
-      if response.statusCode == 503 {
-        throw HTTPClient.parseError(data: responseData, response: response)
+      if response.status.code == 503 {
```
coryan: In these you may be able to say `response.status == .serviceUnavailable`, but I am not sure what is more readable.

packages/storage/Tests/UrlMocks.swift
```diff
@@ -101,131 +131,118 @@ public func expectError<E: Error>(
...
 }
 
 /// A thread-safe registry to store mocks for a specific test run
-public final class MockRegistry: @unchecked Sendable {
-  private static let lock = NSLock()
-  nonisolated(unsafe) private static var registries: [String: MockRegistry] = [:]
+final class MockRegistry: _HTTPClientProtocol, @unchecked Sendable {
+  let id: String
+  private let lock = NSLock()
+  private var mocks: [String: [MockResponse]] = [:]
```
coryan, line 144: FWIW, the "modern Swift" way to do locking is something like:

https://github.com/googleapis/google-cloud-swift/blob/4da340e3f3c0e272df304a61160ae48e47c98491/packages/gax/Tests/MockCredentials.swift#L27

I think if you use that mutex type you can avoid the `@unchecked Sendable`, the compiler can check for you.

For a future PR: I am sure you noticed that the mock credentials use a deque of closures. That allows you to throw exceptions and test the behavior under errors (other than HTTP errors) too.

coryan (approved):
That was fast, thanks!

I think the mocks could use a lot of simplification, but we can do those later.

packages/storage/Tests/UrlMocks.swift
chingor13: Tracking in #423

PR #436, opened 2026-08-17
feat(auth): support API key credentials

coryan:
API key credentials are a simple string that authenticates a service account. They are about as dangerous as they sound, but some services support them, and even recommend them for initial development.

### DIFF
--- a/packages/auth/Sources/GoogleCloudAuth/Providers/ApiKeyCredentials.swift
+++ b/packages/auth/Sources/GoogleCloudAuth/Providers/ApiKeyCredentials.swift
@@ -45,6 +45,6 @@ struct ApiKeyCredentials: CredentialsProvider, Sendable, CustomDebugStringConver
   // MARK: - CustomDebugStringConvertible
 
   var debugDescription: String {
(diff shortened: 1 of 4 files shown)


packages/auth/Sources/GoogleCloudAuth/Providers/ApiKeyCredentials.swift
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  func universeDomain() async -> String? {
+    return nil
+  }
+
+  // MARK: - CustomDebugStringConvertible
+
+  var debugDescription: String {
+    return "ApiKeyCredentials(apiKey: \"[censored]\")"
```
chingor13: nit: "redacted" might be more like other languages and has less connotation

chingor13 (approved):

packages/auth/Sources/GoogleCloudAuth/Providers/ApiKeyCredentials.swift
coryan: Thanks, fixed.

PR #463, opened 2026-08-17
feat(storage)!: return stream of ByteBuffer

### LINKED ISSUE #462: [storage download] should return async stream of data
Currently, we are waiting for the entire content to load and returning a "stream" of a single chunk. By switching to the `AsyncHttpClient` implementation, we should be able to return chunks of data as it's received.

chingor13:
Fixes #462

We now return an iterator of `SwiftNIO.ByteBuffer` rather than `Data`. This makes it so we don't need to make copies of data while converting between types.

### DIFF
--- a/packages/storage/Sources/GoogleCloudStorage/DownloadOptions.swift
+++ b/packages/storage/Sources/GoogleCloudStorage/DownloadOptions.swift
@@ -14,6 +14,7 @@
 
 import Foundation
 @_spi(GoogleCloudInternal) import GoogleCloudGax
+import struct NIOCore.ByteBuffer
 
 /// Specifies a byte range for ranged reads.
 public enum ReadObjectRange: Sendable, Hashable, Equatable {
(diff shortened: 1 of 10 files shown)


packages/storage/Sources/GoogleCloudStorage/DownloadOptions.swift
```diff
@@ -267,27 +273,77 @@ public struct ReadObjectSequence: AsyncSequence, Sendable {
...
-      self.storage = Storage(iterator: iterator)
+    package init(storage: Storage) {
+      self.storage = storage
     }
 
-    /// Advances to the next `Data` chunk in the downloaded object payload stream.
-    public mutating func next() async throws -> Data? {
-      try await storage.iterator.next()
+    /// Advances to the next `ByteBuffer` chunk in the downloaded object payload stream.
+    public mutating func next() async throws -> NIOCore.ByteBuffer? {
```
coryan, line 334: Hmmm... this is sad, if we ever switch the underlying protocol to gRPC + Protobuf, then this would need to change to `Foundation.Data` or we need to perform a copy...

coryan: Created #465 to track this idea....

coryan (approved):
I think we will want a different return type for the stream, but this is an improvement we should merge as-is.

PR #469, opened 2026-08-18
feat(storage)!: retry loop for downloads

chingor13:
Implements retry loop for download

The metadata is returned on the initial request so we want this part of the retry loop. Therefore, we change when the user calls `await`. 

Metadata reading:

```swift
let download = storageClient.readObject(from: "my-bucket", object: "large-file.bin")
let metadata = try await download.metadata
print("Size: \(metadata.size) bytes, Generation: \(metadata.generation)")
```

Data reading:

```swift
let download = storageClient.readObject(from: "my-bucket", object: "large-file.bin")
for try await chunk in download.body {
  processChunk(chunk)
}
```

Since the metadata fetching and data stream are in the same request, we use a coordinator to manage the RPC call.

### DIFF
--- a/packages/storage/Sources/GoogleCloudStorage/DownloadOptions.swift
+++ b/packages/storage/Sources/GoogleCloudStorage/DownloadOptions.swift
@@ -176,6 +176,12 @@ public struct ReadObjectOptions: Sendable {
   /// Flag to enable transparent auto-resumption on transient network failures. Defaults to `true`.
   public var autoResume: Bool = true
 
+  /// Overrides the retry policy for this download.
+  public var retryPolicy: (any RetryPolicy)? = nil
+
+  /// Overrides the backoff policy for this download.
+  public var backoffPolicy: (any BackoffPolicy)? = nil
+
   /// Default configuration options.
   public static var `default`: ReadObjectOptions { ReadObjectOptions() }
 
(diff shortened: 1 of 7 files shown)


packages/storage/Sources/GoogleCloudStorage/DownloadOptions.swift
```diff
@@ -190,6 +196,41 @@ public struct ReadObjectOptions: Sendable {
...
+  bytesReceived: UInt64,
+  totalSize: UInt64?
+) -> ReadObjectRange? {
+  switch originalRange {
+  case .entire:
+    return .fromOffset(bytesReceived)
+  case .fromOffset(let offset):
+    return .fromOffset(offset + bytesReceived)
+  case .prefix(let count):
+    guard count > bytesReceived else { return nil }
```
coryan, line 217: Isn't this an error? The service returning more bytes than requested?

packages/storage/Sources/GoogleCloudStorage/DownloadOptions.swift
```diff
@@ -190,6 +196,41 @@ public struct ReadObjectOptions: Sendable {
...
+    return .bounded(start: bytesReceived, end: count - 1)
+  case .bounded(let start, let end):
+    let newStart = start + bytesReceived
+    guard newStart <= end else { return nil }
```
coryan, line 221: Ditto.

packages/storage/Sources/GoogleCloudStorage/DownloadOptions.swift
```diff
@@ -176,6 +176,12 @@ public struct ReadObjectOptions: Sendable {
   /// Flag to enable transparent auto-resumption on transient network failures. Defaults to `true`.
   public var autoResume: Bool = true
 
+  /// Overrides the retry policy for this download.
+  public var retryPolicy: (any RetryPolicy)? = nil
```
coryan, line 180: A question you should ask yourself: is a retry policy the same as a resume policy? Should we have different types because they answer different questions?

coryan (commented):
Nothing blocking, though maybe the behavior when the service returns more data than expected should be to return an error?

coryan (approved):

packages/storage/Sources/GoogleCloudStorage/DownloadOptions.swift
chingor13: Yes, this is unexpected, but in the context of this method, we probably don't want to throw an exception. We will add.a verification check when we receive the bytes (rather than calculating the recovery range)

packages/storage/Sources/GoogleCloudStorage/DownloadOptions.swift
chingor13: Tracking in #472

PR #504, opened 2026-08-21
refactor(storage): remove unused *.grpc.swift files and dependencies

nolanleastin:


packages/storage/Package.swift
coryan: Does not look like GRPC is used here either.

coryan: And maybe you can remove the `.package(... "swift-grpc")` entry too in that case ?

coryan: Is this used?

coryan (changes requested):
I think you can remove the `GRPC` dependency completely.

packages/storage/Package.swift
nolanleastin: nop

librarian.yaml
```diff
@@ -43,10 +43,6 @@ tools:
       version: 1.38.1
       repo: https://github.com/apple/swift-protobuf.git
       product: protoc-gen-swift
-    - name: protoc-gen-grpc-swift
```
coryan, line 46: The CI is failing because `librarian` is hard-coded to use the grpc-swift plugin, but it is nowhere to be found.

I think you need to:

(1) change librarian to not use `protoc-gen-grpc-swift` if the tool is not configured.
(2) update google-cloud-swift to use that version of librarian
(3) merge this PR

nolanleastin: Merging https://github.com/googleapis/librarian/pull/7361 to go around the problem

coryan (approved):

Issue #16, opened 2026-07-25
test: validate the generated code by running a very long test


coryan:
A number of bugs are hard to detect with simple unit tests. We should run a long test to shake them out, this may catch:

- Leaks: in the form of growing data structures or consuming limited OS resources
- Races and other undefined behavior: Swift makes this hard, but not impossible
- Failures to refresh the access tokens

The idea is to run a process that performs random-ish operations against production.  Run this process until it completes 1 billion operations. If we run this on GCE, it is easy-ish to track memory usage, CPU usage, and other metrics.

The easies service to use is secret manager, it has a rate limit of about 1,000 requests per second (per project). That is about 100 million requests a day, so it may take up to a week to collect all the data.  Allow some time for failed runs and to fix any problems uncovered in this run.

coryan:
After 346,311,033 iterations I got two of these: `io(HTTPClientError.deadlineExceeded)` these are client-side errors, we should be able to recover from them.



```
{
  "insertId": "47vlocf9xr7yx",
  "jsonPayload": {
    "message": "2026-09-06T03:05:53.912128+00:00 dev06 endurance-test[337752]: {\"message\":\"io(HTTPClientError.deadlineExceeded)\",\"labels\":{\"application\":\"endurance-test\",\"version\":\"0.1.0\",\"task\":\"task[8]\"},\"severity\":\"error\"}"
  },
  "resource": {
    "type": "gce_instance",
    "labels": {
      "zone": "us-central1-c",
      "project_id": "coryan-test",
      "instance_id": "6540137710427070891"
    }
  },
  "timestamp": "2026-09-06T03:05:53.975214631Z",
  "labels": {
    "compute.googleapis.com/resource_name": "dev06"
  },
  "logName": "projects/coryan-test/logs/syslog",
  "receiveTimestamp": "2026-09-06T03:05:54.287364178Z"
}
```
https://pantheon.corp.google.com/logs/query;query=timestamp%3D%222026-09-06T03:05:47.976042258Z%22%0AinsertId%3D%2213pixpwfbcswnh%22?project=coryan-test

coryan:
With the fixes to respect the attempt timeout things are looking better... Currently 374,873,745 calls, with no errors, and with 10,537 retries (all successful of course).

```
{
  "insertId": "utomqffg2ne5w",
  "jsonPayload": {
[...]

PR #511, opened 2026-08-22
test: new endurance test

coryan:
Add a manual test to verify things work on long-running applications. As the README explains, running the same program for a long time may uncover bugs such as access token not being refreshed, bad defaults for the retry policies, or slow memory leaks, or crashes under rare scenarios. The test is not guaranteed to find these things, but when it does find them we have real problems.

Towards #16 

LLM disclosure: I used Jetski to translate the [Rust version](https://github.com/googleapis/google-cloud-rust/tree/main/tests/endurance) of this test. And then I read the code and tweaked a couple of things.

### DIFF
--- a/Tests/Endurance/README.md
+++ b/Tests/Endurance/README.md
@@ -1,0 +1,107 @@
+# Google Cloud client libraries for Swift: endurance test
+
+This directory contains a test to verify the client libraries work well in
+long-running applications. Our unit and integration tests are (and should be)
+short-lived. These tests may miss bugs that only manifest themselves when the
+application runs for a long time. Examples of such bugs include:
+
+- Failure to refresh access tokens in the authentication library.
+- Transient errors that appear rarely and are not handled correctly.
+- Race conditions that only appear rarely.
+- Resource leaks, including memory, file descriptors, or any other resource.
+
+While Swift's memory safety and structured concurrency make it hard to
+introduce some of the problems described above, it is not impossible to do so.
+While running a test for a long time does not guarantee that such bugs will be
+found, it makes it less likely that such bugs do exist. In the worst case, such
+a test provides scaffolding to reproduce any bugs reported by our customers.
+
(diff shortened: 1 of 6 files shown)


Tests/Endurance/README.md
```diff
@@ -0,0 +1,115 @@
...
+gcloud projects add-iam-policy-binding ${PROJECT_ID} --role=roles/secretmanager.secretAccessor --member=serviceAccount:${ACCOUNT}
+gcloud projects add-iam-policy-binding ${PROJECT_ID} --role=roles/secretmanager.secretVersionManager --member=serviceAccount:${ACCOUNT}
+```
+
+Create the testing resources:
+
+```shell
+for i in $(seq 0 19); do
+    id=$(printf "secret-%03d" $i)
+    gcloud secrets create --labels=endurance-test=true ${id}
```
chingor13, line 52: Will this error out the script or be an issue if these already exist on the project? What might happen if we accidentally spin up 2 VMs for this test

Tests/Endurance/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Future Work
+
+It would be nice to report metrics, such as successful request counts and
+request latency to Cloud Monitoring.
+
+If we wanted to deploy and run this continuously, it would be nice to use GKE
+and Cloud Build to automatically deploy new versions.
+
+It may be easier to deploy this by building a docker image (there is a Swift
```
chingor13, line 120: Another alternative is to spin up the VM and configure with terraform (via `metadata_startup_script`)

chingor13 (approved):

Tests/Endurance/README.md
coryan: Captured in the README.

Tests/Endurance/README.md
coryan: Captured in the README

PR #524, opened 2026-08-24
impl(gax): use `grpc-swift-2`

### LINKED ISSUE #513: Crash in StorageW1R3 benchmark
This could be a bug in the benchmark, if so, we can change its priority. But if the benchmark is revealing a bug in the storage client, then it is blocking for any release. The bug is a crash, easy to reproduce but takes a couple of minutes:

```
TS=$(date +%s); time swift run --sanitize=address StorageW1R3 --bucket-name coryan-test-us-central1-bm001 --min-object-size 0KiB --max-object-size 64KiB --task-count 256 --iterations 100 2>&1 | tee crash-${TS}.log
Building for debugging...
[0/2] Write swift-version--1F928F6E84187E82.txt
Build of product 'StorageW1R3' complete! (0.36s)
Starting W1R3 benchmark with bucket: coryan-test-us-central1-bm001, tasks: 256, iterations: 100
Generating random payload buffer (65536 bytes)...
Random payload buffer ready.
Counters = ["DELETE_COUNT": 2, "DELETE_ERROR": 0, "READ_COUNT": 393, "READ_ERROR": 0, "SAMPLE_COUNT": 529, "WRITE_COUNT": 134, "WRITE_ERROR": 0]
Counters = ["DELETE_COUNT": 16, "DELETE_ERROR": 0, "READ_COUNT": 1494, "READ_ERROR": 0, "SAMPLE_COUNT": 2018, "WRITE_COUNT": 508, "WRITE_ERROR": 0]
Counters = ["DELETE_COUNT": 30, "DELETE_ERROR": 0, "READ_COUNT": 2627, "READ_ERROR": 0, "SAMPLE_COUNT": 3548, "WRITE_COUNT": 891, "WRITE_ERROR": 0]
Counters = ["DELETE_COUNT": 43, "DELETE_ERROR": 0, "READ_COUNT": 3648, "READ_ERROR": 0, "SAMPLE_COUNT": 4927, "WRITE_COUNT": 1236, "WRITE_ERROR": 0]
Counters = ["DELETE_COUNT": 55, "DELETE_ERROR": 0, "READ_COUNT": 4610, "READ_ERROR": 0, "SAMPLE_COUNT": 6226, "WRITE_COUNT": 1561, "WRITE_ERROR": 0]
Counters = ["DELETE_COUNT": 63, "DELETE_ERROR": 0, "READ_COUNT": 5527, "READ_ERROR": 0, "SAMPLE_COUNT": 7461, "WRITE_COUNT": 1871, "WRITE_ERROR": 0]
Counters = ["DELETE_COUNT": 63, "DELETE_ERROR": 0, "READ_COUNT": 6437, "READ_ERROR": 0, "SAMPLE_COUNT": 8661, "WRITE_COUNT": 2161, "WRITE_ERROR": 1]
Counters = ["DELETE_COUNT": 63, "DELETE_ERROR": 0, "READ_COUNT": 7260, "READ_ERROR": 0, "SAMPLE_COUNT": 9745, "WRITE_COUNT": 2422, "WRITE_ERROR": 2]

*** Signal 4: Backtracing from 0x7657b0ff52c0...swift-backtrace: unable to fetch crash info.
 failed ***


real    1m10.719s
user    16m21.951s
sys     3m42.787s
```

In a prior life, SIGILL (4 - illegal instruction) usually means some memory corruption or undefined behavior. And the address sanitizer is pretty good at catching those, but failed me this time.

I decided to write this down before I lose myself in a debugging session.

### LINKED ISSUE #520: Verify we are using the right version of swift-grpc
Turns out there are two versions of gRPC for Swift:

https://github.com/grpc/grpc-swift
https://github.com/grpc/grpc-swift-2

We are using the first, and yet the second seems to be the modern Swift approach:

https://www.swift.org/blog/grpc-swift-2/

coryan:
Change the implementation to use `grpc-swift-2`, which is the new shinny, designed to use Swift structured concurrency from the bottom up

LLM Disclosure: largely implemented by Jetski.

Fixes #513 and fixes #520

### DIFF
--- a/packages/gax/Sources/GoogleCloudGaxGRPC/GRPCClient.swift
+++ b/packages/gax/Sources/GoogleCloudGaxGRPC/GRPCClient.swift
@@ -13,16 +13,32 @@
 // limitations under the License.
 
 import Foundation
-import GRPC
+import GRPCCore
+import GRPCNIOTransportHTTP2Posix
 import GoogleCloudAuth
 @_spi(GoogleCloudInternal) import GoogleCloudGax
-import NIO
 import SwiftProtobuf
 
 /// Implements a generic gRPC client for the Swift SDK client libraries.
+///
+/// ## Usage and Lifecycle Guidelines
+/// - **Resource Management**: Each `_GRPCClient` manages an underlying HTTP/2 transport connection
+///   loop running inside a background `Task`.
+/// - **Shutdown**:
+///   - Calling ``close()`` initiates a graceful shutdown (`beginGracefulShutdown()`). In-flight RPCs
+///     are allowed to finish executing, while any new RPCs will be rejected.
+///   - If ``close()`` is not explicitly called, the background connection task is automatically shut down
+///     when the `_GRPCClient` instance is deallocated (`deinit`).
+/// - **Async Invariants & Lifetimes**:
+///   - Calls to ``execute(path:request:options:clientHeader:routingParams:)`` retain the `_GRPCClient`
+///     instance for the duration of the asynchronous execution, so the instance will not be deallocated
+///     while an RPC is in flight.
+///   - Calling ``close()`` will allow active in-flight RPCs to drain, but any subsequent calls to
+///     ``execute`` will fail.
 @_spi(GoogleCloudInternal)
-public struct _GRPCClient: Sendable {
-  let connection: ClientConnection
+public final class _GRPCClient: Sendable {
+  let client: GRPCClient<HTTP2ClientTransport.Posix>
+  let connectionTask: Task<Void, any Error>
   let credentials: GoogleCloudAuth.Credentials
 
   public init(from options: ClientOptions, withDefaultEndpoint defaultEndpoint: String) throws {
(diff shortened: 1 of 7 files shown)


packages/gax/Sources/GoogleCloudGaxGRPC/GRPCClient.swift
```diff
@@ -13,16 +13,32 @@
...
+/// - **Async Invariants & Lifetimes**:
+///   - Calls to ``execute(path:request:options:clientHeader:routingParams:)`` retain the `_GRPCClient`
+///     instance for the duration of the asynchronous execution, so the instance will not be deallocated
+///     while an RPC is in flight.
+///   - Calling ``close()`` will allow active in-flight RPCs to drain, but any subsequent calls to
+///     ``execute`` will fail.
 @_spi(GoogleCloudInternal)
-public struct _GRPCClient: Sendable {
-  let connection: ClientConnection
+public final class _GRPCClient: Sendable {
```
chingor13, line 39: non-blocking: we should probably make sure other classes that we don't want to be subclassed are marked `final`.

Tracking in https://github.com/googleapis/google-cloud-swift/issues/525

chingor13 (approved):

packages/gax/Sources/GoogleCloudGaxGRPC/GRPCClient.swift
coryan: Ack. Agree on `final` for classes. Though this is a struct, and you cannot derive from a struct:

> A class can inherit methods, properties, and other characteristics from another **class**. 
https://docs.swift.org/swift-book/documentation/the-swift-programming-language/inheritance

And I think it should remain a `struct`, the owner is a class:

https://github.com/googleapis/google-cloud-swift/blob/c2acae220edb4c6ba94d467bda4debb60aefa397/packages/storage/Sources/GoogleCloudStorage/generated/StorageControl/StorageControl%2BTransport.swift#L30-L31

It does not make sense to make it a class when it will have a single  owner. Did I get something wrong?

PR #526, opened 2026-08-24
feat(storage): introduce ResumePolicy

### LINKED ISSUE #472: [storage] consider creating "resume policy" vs. "retry policy" for uploads/downloads

chingor13:
A `ResumePolicy` is similar to a `RetryPolicy` but is concerned with a series of RPCs which can be recovered/resumed in the case of a failure. The primary use case for now is Storage resumable uploads and downloads.

The `ResumeLoop` is an internal helper for executing a series of RPCs that are resumable. It maintains a `ResumeState` with common loop variables (number of attempts, etc) and can have extra details depending on the workflow.

Fixes #472


packages/gax/Sources/GoogleCloudGax/ResumePolicy.swift
```diff
@@ -0,0 +1,77 @@
...
+  /// Indicates whether the error is transient and recoverable via session resumption.
+  public var isRecoverableForResume: Bool {
+    switch self {
+    case .io:
+      return true
+    case .http(let details):
+      let code = details.http_status_code
+      // 408 (Request Timeout), 429 (Too Many Requests), 502 (Bad Gateway),
+      // 503 (Service Unavailable), 504 (Gateway Timeout) are recoverable.
+      return code == 408 || code == 429 || code == 502 || code == 503 || code == 504
```
coryan: These codes are (I believe ) good for Storage but may not translate to other service can we leave the concrete definition out of `gax` in this iteration?

packages/gax/Sources/GoogleCloudGax/ResumePolicy.swift
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  public func remainingTime(state: ResumeState<Details>) -> Duration? {
+    nil
+  }
+}
+
+extension RequestError {
+  /// Indicates whether the error is transient and recoverable via session resumption.
+  public var isRecoverableForResume: Bool {
```
coryan: Maybe `_isRecoverableForResume` because we don't want application developers to use it?

packages/gax/Sources/GoogleCloudGax/ResumeState.swift
```diff
@@ -0,0 +1,86 @@
...
+// limitations under the License.
+
+import Foundation
+
+/// The input into a resume policy query, tracking progress and error state of an ongoing resumable operation.
+///
+/// On an error or forward progress event, the client library updates and provides an instance of this
+/// class to the ``ResumePolicy``. Specific domains or operations (such as storage uploads and downloads)
+/// provide domain-specific metadata via the `Details` generic parameter.
+public final class ResumeState<Details: Sendable>: @unchecked Sendable {
```
coryan: Why `@unchecked` ?

packages/storage/Sources/GoogleCloudStorage/ResumePolicy.swift
```diff
@@ -0,0 +1,77 @@
...
+// limitations under the License.
+
+import Foundation
+
+/// Determines how errors are handled and when to resume in long-running resumable operations.
+///
+/// Unlike a `RetryPolicy` which operates on a single atomic RPC attempt, a `ResumePolicy` controls
+/// a series of API calls spanning an entire operation (e.g. uploads and downloads). It tracks forward
+/// progress and handles recovery across transient network interruptions, failures, or dropped streams.
+public protocol ResumePolicy<Details>: Sendable {
```
coryan, line 20: I would have started by placing this protocol and types in `GoogleCloudStorage`, with a view to refactor to `gax` when we gained more experience.  It is easier to write a general-purpose policy if you have two or three places where it would be used.

packages/gax/Sources/GoogleCloudGax/StopOnConsecutiveErrors.swift
```diff
@@ -0,0 +1,52 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+/// A ``ResumePolicy`` that permits indefinite resumes across a transfer as long as forward progress
+/// is made, halting only when consecutive errors occur without advancing any bytes.
```
coryan: nit: "any bytes" seems specific ... what if we are trying to resume a download that advances by "rows"?

packages/gax/Sources/GoogleCloudGax/StopOnConsecutiveErrors.swift
```diff
@@ -0,0 +1,52 @@
...
+
+  /// Creates a new `StopOnConsecutiveErrors` instance.
+  ///
+  /// - Parameter maxConsecutiveErrors: The maximum number of consecutive errors without forward progress. Defaults to 3.
+  public init(maxConsecutiveErrors: UInt32 = 3) {
+    self.maxConsecutiveErrors = maxConsecutiveErrors
+  }
+
+  public func onError(state: ResumeState<Details>, error: RequestError) -> ResumeResult {
+    guard error.isRecoverableForResume else {
```
coryan: Consider composition, where an inner policy decides if the error is recoverable. Then we can initialize that inner policy with one type for storage, and a different type for different services.

packages/storage/Sources/GoogleCloudStorage/ResumePolicy.swift
```diff
@@ -0,0 +1,77 @@
...
+  /// The type of domain-specific details associated with this resume policy. Defaults to `Void`.
+  associatedtype Details: Sendable = Void
+
+  /// Queries the policy after an error during an operation.
+  ///
+  /// - Parameters:
+  ///   - state: The current state of the operation.
+  ///   - error: The last error encountered.
+  /// - Returns: The resume decision (`.permanent`, `.exhausted`, or `.resume`).
+  func onError(state: ResumeState<Details>, error: RequestError) -> ResumeResult
```
coryan, line 30: I don't want to complicate things more than we need to... and yet: in Rust we found that some veneers would benefit from using retry policies / loops with a different error type. Should we make this generic on the error type? Of course we would need to introduce aliases that fix the error type for storage (and all the GAPICs).

I will put something in the agenda.

coryan (changes requested):
I don't feel confident about putting this policy in `gax`, but I could be convinced that it will work for other veneers and/or GAPICs.

The stuff around the error type is a longer discussion, I wouldn't block this PR based on that.

packages/gax/Sources/GoogleCloudGax/ResumeState.swift
chingor13: It had to do with mutability of a final class, but we should probably switch to struct instead (similar to the RetryState) type.

packages/gax/Sources/GoogleCloudGax/ResumePolicy.swift
chingor13: refactored to use composition and put the error logic in the storage specific base policy

packages/storage/Sources/GoogleCloudStorage/ResumePolicy.swift
chingor13: Tracking in #537

packages/storage/Sources/GoogleCloudStorage/ResumePolicy.swift
chingor13: Moved to storage. We can move to gax later if something else needs it and we like the interface

packages/storage/Sources/GoogleCloudStorage/ResumeLoop.swift
```diff
@@ -0,0 +1,154 @@
...
+
+import Foundation
+import GoogleCloudGax
+
+/// Runs a resume loop for long-running or multi-step operations (such as resumable uploads and downloads).
+///
+/// Unlike `_RetryLoop` which manages atomic single-RPC retries with `RetryPolicy`, `_ResumeLoop`
+/// tracks forward progress and handles transient error recovery, session resumption,
+/// and backoff across an entire operation using a `ResumePolicy` and `BackoffPolicy`.
+@_spi(GoogleCloudInternal) public struct _ResumeLoop<Details: Sendable>: Sendable {
```
coryan: Since we are only using this in Storage I think we can make it:


```suggestion
struct _ResumeLoop<Details: Sendable>: Sendable {
```

coryan (approved):
Please make `_ResumeLoop` package private and merge. Or if you think it need to remain public let me know why and merge.

PR #542, opened 2026-08-25
fix(storage): data plane returns and handles full bucket names

### LINKED ISSUE #30: [storage] consolidate bucket name references
Should we reference buckets by their full resource name (e.g. `projects/_/buckets/<bucket-name>`, just `<bucket-name>`, or some other alternative.

We should make sure that we use this same type everywhere.

Also note that the HTTP endpoint uses the short bucket name. From https://docs.cloud.google.com/storage/docs/performing-resumable-uploads:

https://storage.googleapis.com/upload/storage/v1/b/BUCKET_NAME/o?uploadType=resumable&name=OBJECT_NAME

### LINKED ISSUE #517: Inconsistent bucket naming between data plane and control plane ops
When we create an object via an upload it returns a `GoogleCloudStorage.Object`. Something like `StorageControlClient.getObject()` or `StorageControlClient.listObjects()` returns the same type. That is all great.

But the value of the `bucket` field is in a different format. The uploads use `<bucket>` where the other operations use `projects/_/bucket/<bucket>`. We should make the data plane operations return buckets in the same format as they are used elsewhere.

chingor13:
Fixes #517 
Fixes #30

### DIFF
--- a/packages/storage/Sources/GoogleCloudStorage/BucketName.swift
+++ b/packages/storage/Sources/GoogleCloudStorage/BucketName.swift
@@ -22,22 +22,14 @@ package final class BucketName: Sendable {
   /// (`projects/_/buckets/<bucket>`).
   ///
   /// If `bucket` is empty, this returns an empty string. If `bucket` is already in a
-  /// `projects/...` format, it is returned unmodified. Otherwise, `projects/_/buckets/`
+  /// `projects/_/buckets/` format, it is returned unmodified. Otherwise, `projects/_/buckets/`
   /// is prepended.
   package static func formatResourceName(_ bucket: String) -> String {
     guard !bucket.isEmpty else { return "" }
-    if bucket.hasPrefix("projects/") {
+    if bucket.hasPrefix("projects/_/buckets/") {
       return bucket
     }
     return "projects/_/buckets/\(bucket)"
-  }
-
-  /// Formats a project and bucket name into the canonical resource name format
-  /// (`projects/<project>/buckets/<bucket>`).
-  package static func format(project: String = "_", bucket: String) -> String {
-    guard !bucket.isEmpty else { return "" }
-    let proj = project.isEmpty ? "_" : project
-    return "projects/\(proj)/buckets/\(bucket)"
   }
 
   /// Alias for `formatResourceName(_:)`.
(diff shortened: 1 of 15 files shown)


packages/storage/Sources/GoogleCloudStorage/BucketName.swift
```diff
@@ -0,0 +1,87 @@
...
+    return "projects/_/buckets/\(bucket)"
+  }
+
+  /// Formats a project and bucket name into the canonical resource name format
+  /// (`projects/<project>/buckets/<bucket>`).
+  public static func format(project: String = "_", bucket: String) -> String {
+    guard !bucket.isEmpty else { return "" }
+    let proj = project.isEmpty ? "_" : project
+    return "projects/\(proj)/buckets/\(bucket)"
+  }
```
coryan: AFAIK, project-scoped buckets are not a thing. They may never be.  Obviously the full name was chosen to support them someday, but I do not believe any projects exist to implement them.  should make this function private because it may lead people to believe they exist and get confused.

packages/storage/Sources/GoogleCloudStorage/StorageClient+Download.swift
```diff
@@ -169,8 +169,9 @@ extension GoogleCloudGax._HTTPClient {
       CharacterSet(charactersIn: "/"))
     let encodedObject =
       object.addingPercentEncoding(withAllowedCharacters: allowedObjectCharacters) ?? object
+    let bucketId = BucketName.extractBucketName(bucket)
```
coryan, line 172: This is more generous than the approach we took in Rust.  There we made it an error to provide just "my-bucket" instead of "projects/_/buckets/my-bucket".  I think the norm in Swift is to be more forgiving, but maybe do some research, or ask the all-knowning-LLM-overlords for an opinion.

coryan (approved):
Please hide the project-scoped bucket alternative, that is confusing.

The thing about how Rust is more strict is just FYI. I trust your decision here.

packages/storage/Sources/GoogleCloudStorage/StorageClient+Download.swift
chingor13: If looks like in Python, Go, Java, Node, C++, C#, Ruby, and PHP, users specify the bucket name with just the plain name.

Issue #450, opened 2026-04-09
librarian: support the Swift releases


coryan:
Most likely, Swift will need split repos for releases: all the development will happen in a monorepo, each package will have a repo that gets code only when a new release is needed.

- [x] We will need librarian to support this workflow.
- [x] We need the right policies and exemptions to support this workflow.
- [x] update swift-google-rpc usage
- [x] release swift-google-gax
- [x] update swift-google-gax, swift-google-type, swift-google-cloud-common, swift-google-api usage
- [x] release swift-google-longrunning
- [x] update swift-google-longrunning usage
- [x] release swift-google-cloud-storage

chingor13:
We have a design for how to create the separate repositories in the googleapis organization. [go/cloud-swift:split-repo-creation](http://goto.google.com/cloud-swift:split-repo-creation)

chingor13:
235 split repositories are all created via terraform

PR #545, opened 2026-08-26
deps: update wkt, auth dependencies to remote url

chingor13:
Towards #450

Also, updates the CI script to be able to use local code for testing. Example found failure: in #618

### DIFF
--- a/ci/package-dependencies.sh
+++ b/ci/package-dependencies.sh
@@ -23,17 +23,28 @@ _REMOVED_DISABLE_RESOLUTION=()
 
 edit_package_dependencies() {
     local dir="$1"
+    local clean_dir="${dir#./}"
+    [[ -z "${clean_dir}" ]] && clean_dir="."
     _EDITED_PACKAGES+=("${dir}")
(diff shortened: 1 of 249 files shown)


ci/package-dependencies.sh
```diff
@@ -0,0 +1,109 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
+
+_EDITED_PACKAGES=()
+_REMOVED_DISABLE_RESOLUTION=()
+
+edit_package_dependencies() {
```
coryan, line 23: We are close to the "this is a tool" territory.  Fine for now, but if it gets more complicated or we find ourselves changing these often then we may need to write a tool in Swift or Go.

coryan (changes requested):
I think we need to update the `docs/contributor/*.md` guides and the GEMINI.md  to talk about these `swift package edit` commands that one needs to run?

coryan (approved):
> I think we need to update the docs/contributor/*.md guides and the GEMINI.md to talk about these swift package edit commands that one needs to run?

On second thought, you can make those changes in a follow up PR. But maybe tell us the secrets in a brief note or message to the chat room.

PR #547, opened 2026-08-27
[WIP] refactor(storage): use observer for tracking upload/download progress

### LINKED ISSUE #28: [storage upload] reconsider how to inject observability into upload progress
Address this comment:

----

This is interesting.... An alternative might have been to wrap the `UploadSource` into an `UploadSourceWithStatus` ... That way only applications that care about the progress can choose how the progress is notified (a var, a callback, some OTel counter, etc.)

chingor13:
## Overview

This PR introduces an observer-based observability architecture for Google Cloud Storage uploads and downloads via a generic `OperationObserver<Context, Progress, Result>` protocol, replacing the previous `AsyncStream<UploadStatus>` model.

## Motivation & Background

Previously, monitoring upload progress required calling `task.makeStatusStream()` and spawning an independent concurrent `Task` to consume progress events while simultaneously awaiting `task.value`. This model had several limitations:
- **Complex Task Orchestration**: Callers had to manage multiple concurrent tasks (one consuming the stream, one awaiting the result).
- **Limited Observability**: The stream only reported coarse progress numbers and final completion/cancellation, lacking visibility into transient error retries, backoff delays, or per-chunk timing.
- **Upload-Only**: The stream mechanism was coupled to uploads and could not be reused across other operations like streaming downloads.

## Key Changes

1. **Generic Core Observer (`OperationObserver.swift`)**:
   - Defined `OperationObserver<Context, Progress, Result>: Sendable` with default no-op implementations for lifecycle and resilience callbacks:
     - `operationDidStart(context: Context)`
     - `progressUpdated(_ progress: Progress)`
     - `operationDidRetry(_ retry: RetryDetails)`
     - `operationDidComplete(result: Result, totalDuration: Duration)`
     - `operationDidFail(error: any Error)`

2. **Domain Specializations (`UploadObserver.swift` & `DownloadObserver.swift`)**:
   - `UploadObserver`: Specializes `OperationObserver` for `StorageOperationContext`, `UploadProgress` (`TransferProgress`), and `Object`, adding `chunkDidComplete(index:byteRange:duration:)`.
   - `DownloadObserver`: Specializes `OperationObserver` for `StorageOperationContext`, `DownloadProgress` (`TransferProgress`), and `ReadObjectMetadata`, adding `chunkDidReceive(bytes:totalReceived:)`.
   - Composite dispatchers (`_CompositeUploadObserver`, `_CompositeDownloadObserver`) enable chaining multiple observers.

3. **Storage Operation Models (`StorageOperation.swift`)**:
   - `StorageOperationContext`: Context struct containing `bucket`, `object`, and optional `sessionId` (e.g. upload URI).
   - `TransferProgress`: Represents transferred bytes, total size, and fraction completed with convenience properties for uploads (`bytesUploaded`) and downloads (`bytesDownloaded`).

4. **Removed `UploadTask`**:
   - We no longer need to have 2 awaitable types (metadata and status stream)

5. **Client-Level and Per-Request Registration**:
   - Added `observers: [any UploadObserver]` to `UploadOptions` and `StorageClientOptions.upload`.
   - Added `observers: [any DownloadObserver]` to `ReadObjectOptions` and `StorageClientOptions.download`.

---

### Before (Stream-based monitoring)

```swift
let task = client.upload(fileSource, to: "my-bucket", as: "large-file.bin")

// Required spawning a separate task to consume the status stream concurrently
let streamTask = Task {
  for await status in task.makeStatusStream() {
    switch status {
    case .inProgress(let progress):
      print("Uploaded $progress.bytesUploaded) bytes")
    case .completed(let object):
      print("Upload completed: $object.name)")
    case .failed(let error):
      print("Upload failed: $error)")
    case .cancelled:
      print("Upload cancelled")
    }
  }
}

// Await the upload result
let object = try await task.value
streamTask.cancel()
```

### After (Observer-based monitoring)

```swift
// 1. Define an observer (implement only the callbacks you need)
struct ProgressLoggingObserver: UploadObserver {
  func operationDidStart(context: StorageOperationContext) {
    print("Started upload for \(context.bucket)/\(context.object)")
  }

  func progressUpdated(_ progress: UploadProgress) {
    if let percent = progress.percentCompleted {
      print("Upload progress: \(progress.bytesUploaded)/\(progress.totalBytes ?? 0) bytes (\(String(format: "%.1f", percent))%)")
    } else {
      print("Upload progress: \(progress.bytesUploaded) bytes")
    }
  }

  func operationDidRetry(_ retry: RetryDetails) {
    print("Warning: Upload attempt \(retry.attempt) failed (\(retry.error)). Retrying in \(retry.backoff)...")
  }

  func operationDidComplete(result: Object, totalDuration: Duration) {
    print("Upload finished in \(totalDuration). Object size: \(result.size) bytes.")
  }

  func operationDidFail(error: any Error) {
    print("Upload permanently failed: \(error)")
  }
}

// 2. Attach observer via options (or client-level defaults)
let options = UploadOptions().with {
  $0.addObserver(ProgressLoggingObserver())
}

// 3. Simple, direct await without secondary task orchestration
let object = try await client.upload(fileSource, to: "my-bucket", as: "large-file.bin", options: options)
```

Fixes #28


coryan (approved):
FWIW, I think this is a lot of code for a feature that few people would need.  And if they really need it, they have other ways to get information about progress:

- If they really want to know how much data is uploaded, they can provider their own data source where they observe the bytes requested by the library.
- If they want to monitor RPCs and attempts, they can provider their own resume policies where they observe the status of several calls.


Not as direct as this approach, but enough information to troubleshoot problems.

I am going to let you make the call on this, my advice is to not implement this feature until we have somebody asking for it.

chingor13:
> FWIW, I think this is a lot of code for a feature that few people would need. And if they really need it, they have other ways to get information about progress:
> 
> * If they really want to know how much data is uploaded, they can provider their own data source where they observe the bytes requested by the library.
> * If they want to monitor RPCs and attempts, they can provider their own resume policies where they observe the status of several calls.
> 
> Not as direct as this approach, but enough information to troubleshoot problems.
> 
> I am going to let you make the call on this, my advice is to not implement this feature until we have somebody asking for it.

I think we'll hold off on the observer pattern, but simplify the return type of `client.upload`. Having that return type have a status stream doesn't feel like the right design for observability. If users need it, we can bring back this fuller implementation with the observer pattern.

chingor13:
Closing in favor of #549

PR #550, opened 2026-08-27
fix(storage)!: use AnyByteStorage as return type for storage downloads

### LINKED ISSUE #465: storage: consider returning a wrapper type for the data
The ideal data type to return for SwiftNIO-based downloads is `NIOCore.ByteBuffer`, that makes zero copies. The ideal data type to return for gRPC-based downloads (such as bidirectional streaming reads aka `openObject` calls) is a `Foundation.Data`.

Unfortunately we cannot convert between the two without copying the bytes.

That suggests we need a wrapper type that can represent either, something along the lines of:

```swift
import Foundation
import NIOCore

// It should be a struct that contains an enum, because we don't want the enum to be public...
public enum AnyByteStorage: Sendable {
    case data(Data)
    case byteBuffer(ByteBuffer)
}

extension AnyByteStorage {
    public var count: Int {
        switch self {
        case .data(let data):
            return data.count
        case .byteBuffer(let buffer):
            return buffer.readableBytes
        }
    }

    public var isEmpty: Bool {
        return count == 0
    }

    public func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R {
        switch self {
        case .data(let data):
            return try data.withUnsafeBytes(body)
        case .byteBuffer(let buffer):
            return try buffer.withUnsafeReadableBytes(body)
        }
    }

    public var byteArray: [UInt8] {
        withUnsafeBytes { Array($0) }
    }
}
```

chingor13:
Fixes #465

This allows us to own the return type for storage (hides the implementation detail of SwiftNIO.ByteBuffer) and also allow for interop with `Foundation.Data` return types.


packages/swift-google-cloud-storage/Sources/GoogleCloudStorage/DownloadOptions.swift
```diff
@@ -342,9 +342,9 @@ public struct ReadObjectMetadata: Sendable, Hashable, Equatable {
   }
 }
 
-/// An asynchronous sequence of `ByteBuffer` chunks representing an object payload being downloaded.
+/// An asynchronous sequence of `AnyByteStorage` chunks representing an object payload being downloaded.
 public struct ReadObjectSequence: AsyncSequence, Sendable {
-  public typealias Element = NIOCore.ByteBuffer
+  public typealias Element = AnyByteStorage
```
coryan: The name is hard to grok, is it a think to store any bytes? or to store anys as bytes? Just `StorageData`? or `ByteBuffer`?

packages/swift-google-gax/Sources/GoogleCloudGax/AnyByteStorage.swift
```diff
@@ -0,0 +1,202 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+import NIOCore
+import NIOFoundationCompat
+
+/// A type-erased container representing a sequence of bytes backed either by `Foundation.Data`
+/// or `NIOCore.ByteBuffer` without unnecessary memory copying.
+public struct AnyByteStorage: Sendable {
```
coryan: If we refactor this to `gax`, then all clients have to compile it, even if only `Storage` uses it. That seems like a bad tradeoff.

Unless you see a clear reason to refactor now, I would leave it on the storage library.

coryan (approved):
The naming bikeshed is optional, but please it some thought and decide.

The location less so. Maybe you have a use-case in mind already for this thing in `gax`, if not, please keep it in storage.

PR #558, opened 2026-08-31
perf(storage): avoid making extra Data copies in simple upload flow

### LINKED ISSUE #29: [storage upload]: ensure we aren't copying extra data
Address this comment:

----
Referencing the code in StorageClient+Upload.swift:

>     body.append("--\(boundary)\r\n".data(using: .utf8)!)
>    body.append("Content-Type: application/octet-stream\r\n\r\n".data(using: .utf8)!)
>    body.append(data)

Do you know if this makes a data copy? Ideally we stream the data without having to copy several MiB each time. Feel free to leave that to a follow up PR.

chingor13:
Fixes #29

### DIFF
--- a/packages/swift-google-cloud-storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
+++ b/packages/swift-google-cloud-storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
@@ -755,9 +755,9 @@ extension StorageClient {
     var buffer = ByteBufferAllocator().buffer(capacity: totalCapacity)
 
     buffer.writeString(preamble)
-    metadataJson.withUnsafeBytes { buffer.writeBytes($0) }
+    _ = metadataJson.withUnsafeBytes { buffer.writeBytes($0) }
     buffer.writeString(middle)
-    data.withUnsafeBytes { buffer.writeBytes($0) }
+    _ = data.withUnsafeBytes { buffer.writeBytes($0) }
     buffer.writeString(epilogue)
 
     request.setBody(buffer: buffer)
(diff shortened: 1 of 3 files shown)


packages/swift-google-gax/Sources/GoogleCloudGax/HTTPClientRequest.swift
```diff
@@ -52,11 +61,20 @@ import struct Logging.Logger
...
 
   public mutating func setBody(data: Data, ofContentType: String) {
-    self.body = data
+    self.body = .data(data)
+    self.headers.replaceOrAdd(name: "Content-Type", value: ofContentType)
+  }
+
+  public mutating func setBody(buffer: NIOCore.ByteBuffer) {
+    self.body = .byteBuffer(buffer)
+  }
```
coryan, line 74: I wonder if we could use this:

https://swiftpackageindex.com/swift-server/async-http-client/1.36.0/documentation/asynchttpclient/httpclientrequest/body-swift.struct/init(length:startupload:)

To initialize the multi-part upload from a `[ByteBuffer]` ... and then return each "part" separately, without copying, including the payload....

packages/swift-google-cloud-storage/Sources/GoogleCloudStorage/StorageClient+Upload.swift
```diff
@@ -740,22 +742,25 @@ extension StorageClient {
...
+    var buffer = ByteBufferAllocator().buffer(capacity: totalCapacity)
+
+    buffer.writeString(preamble)
+    _ = metadataJson.withUnsafeBytes { buffer.writeBytes($0) }
+    buffer.writeString(middle)
+    _ = data.withUnsafeBytes { buffer.writeBytes($0) }
+    buffer.writeString(epilogue)
 
-    request.setBody(data: body)
+    request.setBody(buffer: buffer)
```
coryan, line 763: meaning here we would provide something like:

```
request.setBody(parts: [preamble.utf8, metadataJson.utf8, middle.utf8, buffer, epilogue.utf8])
```

And then we avoid all copying (other than the copying into `buffer` but maybe we can fix that in a second pass?

packages/swift-google-gax/Sources/GoogleCloudGax/HTTPClientRequest.swift
chingor13: I'm looking into an alternative in #565 that builds the multipart body as a stream

PR #560, opened 2026-08-31
impl: framework for storage samples

### LINKED ISSUE #556: Write a framework to compile the sample code

coryan:
Create a framework to populate the storage samples. Things that I like about this structure include:

- Each sample in its own file, so we can show the imports and the function parameters.
- All the samples are in one product, so they are compiled and linked only once.
- The samples can be part of the integration tests.
- Cleaning up buckets is in a single place.
- The structure is somewhat similar to Rust, so maybe the LLMs can translate.

Things I don't like:

- The logging is not awesome.
- Cleaning up stale buckets may take multiple days.

Fixes #556

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -149,6 +149,24 @@ let package = Package(
       path: "Tests/StorageW1R3",
       exclude: ["README.md"]
     ),
+    .target(
+      name: "StorageSamples",
+      dependencies: [
+        .product(name: "GoogleCloudStorage", package: "swift-google-cloud-storage"),
+        .product(name: "GoogleCloudAuth", package: "swift-google-auth"),
+        .product(name: "GoogleCloudGax", package: "swift-google-gax"),
+        .product(name: "Logging", package: "swift-log"),
+      ],
+    ),
+    .testTarget(
+      name: "StorageSamplesDriver",
+      dependencies: [
+        "StorageSamples",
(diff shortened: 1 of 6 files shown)


chingor13 (changes requested):
I don't see what's setting the integration test mark label when the bucket is first created.

It looks like the cleanup is listing all buckets, then filtering to ones with the label. Then it updates those buckets to mark requester pays and add set the test mark label.

coryan:
> I don't see what's setting the integration test mark label when the bucket is first created.

The example is not supposed to show the ugly labels we use for our own cleanup.  So after the test we call `cleanupTestBuckets()` which calls `markTestBuckets()` and sets the label.

> It looks like the cleanup

Not "the" cleanup. The last resort cleanup... the normal path is that the test cleans up after itself.  The `cleanupStaleTestBuckets()` is a last resort if that fails...

>  is listing all buckets, then filtering to ones with the label. Then it updates those buckets to mark requester pays and add set the test mark label.

Yes... the idea is that some buckets may leak even after calling `cleanupTestBuckets()` after the test(s). We will have objects with retention periods (a few days) that cannot be deleted during the test and therefore their buckets cannot be deleted either. We will have tests crash while trying to delete things. The `cleanupStaleTestBuckets()` function is a last resort to safely delete those.

I added some comments and removed some duplicate requests. PTAL.

chingor13 (approved):

PR #565, opened 2026-08-31
perf(storage): use streaming for building multipart upload request for simple uploads

### LINKED ISSUE #29: [storage upload]: ensure we aren't copying extra data

### LINKED ISSUE #540: Optimization idea for Storage client
When performing simple uploads we copy all the data into a single buffer and then send the data via a single `POST`.

We could save ourselves some data copying by adapting the source stream and passing it directly to the HTTP client, it could then send the stream without any data copies (or with minimal copies).

chingor13:
refactor(storage): UploadSource returns ByteBuffer chunks

This improves performance of simple uploads by having the body of a multipart upload built on-the-fly rather than reading the entire body into a Data struct and appending. Instead, we wrap using NIOCore.ByteBuffer to buffer the contents.

Fixes #29 
Fixes #540


chingor13:
I am going to break this into smaller PRs

chingor13:
Closing in favor of #601

PR #592, opened 2026-09-01
docs(storage): create bucket samples

### LINKED ISSUE #590: [storage]: write the create bucket samples
- [ ] Write the `storage_create_bucket_class_location` sample
- [ ] Write the `storage_create_bucket_dual_region` sample
- [ ] Write the `storage_create_bucket_hierarchical_namespace` sample
- [ ] Write the `storage_create_bucket_turbo_replication` sample
- [ ] Write the `storage_create_bucket_with_encryption_enforcement` sample
- [ ] Write the `storage_create_bucket_with_object_retention` sample

coryan:
Implement multiple samples for `createBucket()`. The service offers several configuration options, and they like to write a sample for (almost) each option.

In general, I chose to hard-code things like the location, or the storage class. That makes the samples far easier to grok, and the cost for us (fixing the strings) is almost zero as the key locations have not changed in a decade or more.

Fixes #590

### DIFF
--- a/Sources/StorageSamples/Buckets/CreateBucketHierarchicalNamespace.swift
+++ b/Sources/StorageSamples/Buckets/CreateBucketHierarchicalNamespace.swift
@@ -1,0 +1,41 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START storage_create_bucket_hierarchical_namespace]
+import GoogleCloudStorage
+
+public func createBucketHierarchicalNamespace(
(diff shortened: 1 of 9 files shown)


Sources/StorageSamples/Buckets/CreateBucketHierarchicalNamespace.swift
```diff
@@ -0,0 +1,41 @@
...
+          $0.project = "projects/\(projectId)"
+          $0.hierarchicalNamespace = .init().with {
+            $0.enabled = true
+          }
+          $0.iamConfig = .init().with {
+            $0.uniformBucketLevelAccess = .init().with {
+              $0.enabled = true
+            }
+          }
+        }
```
chingor13, line 37: non-blocking: For these nested `.init().with` blocks, do we want to use naming for the inner context?

```swift
$0.bucket = .init().with { bucket in
  bucket.project = "projects/\(projectId)"
  bucket.hierarchicalNamespace = .init().with { ns in
    ns.enabled = true
  }
  bucket.iamConfig = .init().with { iam in
    iam.uniformBucketLevelAccess = .init().with { access in
      access.enabled = true
    }
  }
}
```

chingor13 (approved):

Sources/StorageSamples/Buckets/CreateBucketHierarchicalNamespace.swift
coryan: I think we should name the inner context(s). Done.

chingor13 (approved):

PR #593, opened 2026-09-01
feat(gax): allow providing AsyncSequence as stream for HttpClient

### LINKED ISSUE #572: [storage upload] HttpClient should be able to handle a stream of data
This will allow us to provide a raw stream without buffering or copying data

chingor13:
Fixes #572

### DIFF
--- a/packages/swift-google-gax/Sources/GoogleCloudGax/HTTPClientRequest.swift
+++ b/packages/swift-google-gax/Sources/GoogleCloudGax/HTTPClientRequest.swift
@@ -92,33 +92,41 @@ enum _RequestBody: Sendable, Equatable {
 
   public mutating func setBody<S: AsyncSequence & Sendable>(
     stream: S,
-    length: AsyncHTTPClient.HTTPClientRequest.Body.Length = .unknown
+    length: Int64? = nil
   ) where S.Element == NIOCore.ByteBuffer {
-    self.body = .stream(.stream(stream, length: length))
+    let bodyLength: AsyncHTTPClient.HTTPClientRequest.Body.Length =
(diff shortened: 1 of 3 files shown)


packages/swift-google-gax/Sources/GoogleCloudGax/HTTPClientRequest.swift
```diff
@@ -78,6 +79,46 @@ enum _RequestBody: Sendable, Equatable {
     self.headers.replaceOrAdd(name: "Content-Type", value: ofContentType)
   }
 
+  public mutating func setBody<S: AsyncSequence & Sendable>(
+    stream: S,
+    length: Int64? = nil
+  ) where S.Element == NIOCore.ByteBuffer {
```
coryan: nit: splitting the constraint like this seems odd to me, consider:

```suggestion
  public mutating func setBody<S>(stream: S, length: Int64? = nil)
  where S: AsyncSequence & Sendable, S.Element == NIOCore.ByteBuffer {
```

I am not 100% sure of the correct syntax, but you get the idea.

packages/swift-google-gax/Tests/HTTPClientV2Tests.swift
coryan, line 1: unrelated: We can probably rename this test file since we no longer have a `HttpClientTests.swift` file to conflict with it.

coryan (approved):

packages/swift-google-gax/Tests/HTTPClientV2Tests.swift
chingor13: Tracking in #596

PR #605, opened 2026-09-02
perf(storage): ByteBuffer.subdata returns a zero-copy sub-buffer

### LINKED ISSUE #604: [storage upload] BytesSource copies when using Foundation.Data type
`Foundation.Data.subdata()` returns a copy of the data in the requested range. We should avoid this if possible.

chingor13:
Fixes #604 

Foundation.Data can return the same heap bytes using slice/index rather than subdata.


coryan (approved):

coryan:
I think I am starting to like ByteBuffer

PR #607, opened 2026-09-02
perf(storage): maintain file descriptor to avoid multiple file opens

### LINKED ISSUE #606: [storage upload] FileSource opens file multiple times
We are opening the file multiple times and seeking to a spot rather than keeping a file pointer

chingor13:
Fixes #606 

We use a box implementation to manage the lifecycle of the file descriptor (close on deinit if it was opened)

### DIFF
--- a/packages/swift-google-cloud-storage/Sources/GoogleCloudStorage/FileSource.swift
+++ b/packages/swift-google-cloud-storage/Sources/GoogleCloudStorage/FileSource.swift
@@ -14,6 +14,7 @@
 
 import Foundation
 import NIOCore
+// We expect the API is stable, see https://github.com/apple/swift-nio/issues/3052 for details
 import _NIOFileSystem
 
 private final class FileHandleBox: @unchecked Sendable {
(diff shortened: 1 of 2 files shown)


packages/swift-google-cloud-storage/Sources/GoogleCloudStorage/FileSource.swift
```diff
@@ -13,13 +13,106 @@
...
 import Foundation
+import NIOCore
+
+#if canImport(Darwin)
+  import Darwin
+#elseif canImport(Glibc)
+  import Glibc
+#elseif canImport(Musl)
+  import Musl
+#endif
```
coryan: What about Windows?  Allegedly we want to support that too (we are missing a CI build for it).

Maybe:


```suggestion
#elseif canImport(Musl)
  import Musl
#else
  #error("unsupported platform, please file a bug at ...")
#endif
```

packages/swift-google-cloud-storage/Sources/GoogleCloudStorage/FileSource.swift
```diff
@@ -13,11 +13,51 @@
 // limitations under the License.
 
 import Foundation
+import NIOCore
+import _NIOFileSystem
```
coryan, line 18: Maybe leave a comment explaining why we took a dep on a (in principle) unstable API.


```suggestion
// We expect the API is stable, see https://github.com/apple/swift-nio/issues/3052 for details
import _NIOFileSystem
```

coryan (approved):
Can we make the NIO

Issue #647, opened 2026-09-04
Run the `StorageW1R3` benchmark and capture the results.


coryan:
We should run the storage benchmark in following configurations:

- [x] Fixed object size at 8KIB, no more than one task per CPU
- [x] Fixed object size at 2MiB, no more than one task per CPU
- [x] Fixed object size at 128MiB, no more than one task per CPU
- [x] Random object size from 0KiB to 256MiB, about 2 tasks per CPU

The last is more a torture test than a real benchmark. We want to verify there are no crashes but at that concurrency level we won't get the best throughput (the NIC may saturate before anything else).

coryan:
While running the benchmark I got errors that look funky:

```
123,1507,13743851133,SINGLE_SHOT,215756484,0,11976287,fdXJSRo0uSpnzvutwQ8uyc6XnNJjNZLe,true,ERR,DELETE_COUNT=30295;DELETE_ERROR=0;READ_COUNT=1829700;READ_ERROR=0;SAMPLE_COUNT=2470094;WRITE_COUNT=610099;WRITE_ERROR=0;TASK_COUNT=384;error=service(GoogleCloudGax.ServiceError(code: GoogleRpc.Code.unknown; message: "We encountered an internal error. Please try again."; details: []))
156,8710,70370641771,SINGLE_SHOT,57140918,0,10844337,k79Qyed5gZMjez0RBgwJnlmdTZwJcruH,true,ERR,DELETE_COUNT=157507;DELETE_ERROR=0;READ_COUNT=9461081;READ_ERROR=0;SAMPLE_COUNT=12772479;WRITE_COUNT=3153891;WRITE_ERROR=1;TASK_COUNT=384;error=service(GoogleCloudGax.ServiceError(code: GoogleRpc.Code.unknown; message: "We encountered an internal error. Please try again."; details: []))
235,8840,71278354912,SINGLE_SHOT,41082906,0,10738937,VbDa0tTy3KqQfpuLuWJDVl2iA7ZINPZY,false,ERR,DELETE_COUNT=159533;DELETE_ERROR=0;READ_COUNT=9583042;READ_ERROR=0;SAMPLE_COUNT=12937117;WRITE_COUNT=3194542;WRITE_ERROR=2;TASK_COUNT=384;error=service(GoogleCloudGax.ServiceError(code: GoogleRpc.Code.unknown; message: "We encountered an internal error. Please try again."; details: []))
```

[...]

PR #649, opened 2026-09-04
feat(gax): support explicit client shutdown

coryan:
Add explicit shutdown() to _HTTPClientProtocol, HTTPClientHolder, and _HTTPClient for deterministic resource cleanup.

Motived by #647

### DIFF
--- a/packages/swift-google-gax/Sources/GoogleCloudGax/HTTPClientHolder.swift
+++ b/packages/swift-google-gax/Sources/GoogleCloudGax/HTTPClientHolder.swift
@@ -13,38 +13,67 @@
 // limitations under the License.
 
 import class AsyncHTTPClient.HTTPClient
+import struct AsyncHTTPClient.HTTPClientError
 import struct AsyncHTTPClient.HTTPClientRequest
 import struct AsyncHTTPClient.HTTPClientResponse
 import struct Logging.Logger
+import Synchronization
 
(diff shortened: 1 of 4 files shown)


packages/swift-google-gax/Sources/GoogleCloudGax/HTTPClientHolder.swift
```diff
@@ -13,38 +13,67 @@
...
     }
   }
 
+  func shutdown() async throws {
+    let client = self.state.withLock { state -> AsyncHTTPClient.HTTPClient? in
+      let client = state
+      state = nil
+      return client
+    }
+    if let client {
```
chingor13, line 67: This (second/third) `let` looks odd to me, but maybe it's actually correct.

chingor13 (approved):

packages/swift-google-gax/Sources/GoogleCloudGax/HTTPClientHolder.swift
coryan: FWIW, I think it is "correct", but after thinking about it, I think the invariants assumed in `deinit{}` may no longer hold, so this change looks more risky now.  I am pondering...

coryan:
I am going to close this for now.  If we every need it, we can introduce shutdowns back.

PR #695, opened 2026-09-04
deps: use remote swift-google-rpc repo

chingor13:
Towards #450

### DIFF
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -145,7 +145,8 @@ default:
         path: generated/swift-google-longrunning
         api_package: google.longrunning
       - name: GoogleRpc
-        path: generated/swift-google-rpc
+        url: https://github.com/googleapis/swift-google-rpc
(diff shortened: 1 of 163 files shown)


librarian.yaml
```diff
@@ -145,7 +145,8 @@ default:
         path: generated/swift-google-longrunning
         api_package: google.longrunning
       - name: GoogleRpc
-        path: generated/swift-google-rpc
+        url: https://github.com/googleapis/swift-google-rpc
+        version: 0.1.0-preview
```
coryan, line 149: Food for thought: It would be nice if we did not have to repeat the version here... Librarian could learn to pick the latest version of maybe we just use `1.0` always (once we get there).

coryan (approved):

librarian.yaml
chingor13: Tracking in #696

Issue #594, opened 2026-09-01
[storage]: error deserializing `GoogleLongRunning.Operation`


coryan:
While trying to write the `renameFolder()` sample I got this output:

```
running renameFolder() sample
DEBUG DEBUG converting Operation from: StorageControlProtos.Google_Longrunning_Operation:
name: "projects/_/buckets/swift-sdk-testing-ga8f8azhoey8k9evxc7v8bgv18nx8a0g9zvd77zjc9mwv/operations/CiQ0MmIwMDRkYi1jM2EwLTQ0MjMtODcwMi0wZGQ3YzEwNjEzNGIQAQ"
metadata {
  type_url: "type.googleapis.com/google.storage.control.v2.RenameFolderMetadata"
  value: "\n;\n\f\b\260\324\334\324\006\020\240\237\353\200\002\022\f\b\260\324\334\324\006\020\240\237\353\200\002\032\f\b\260\324\334\324\006\020\240\237\353\200\002\"\rrename-folder0d\022\022example-folder-id/\032\022renamed-folder-id/"
}
done: true
response {
  type_url: "type.googleapis.com/google.storage.control.v2.Folder"
  value: "\nmprojects/_/buckets/swift-sdk-testing-ga8f8azhoey8k9evxc7v8bgv18nx8a0g9zvd77zjc9mwv/folders/renamed-folder-id/\030\001\"\f\b\257\324\334\324\006\020\230\327\310\214\003*\v\b\260\324\334\324\006\020\256\222\202|"
}

DEBUG DEBUG cannot convert Operation.metadata field: anyTranscodeFailure
DEBUG DEBUG error in renameFolder()/Operation: anyTranscodeFailure
✘ Test runFolderSamples() recorded an issue at Driver.swift:39:4: Caught error: anyTranscodeFailure
✘ Test runFolderSamples() failed after 3.443 seconds with 1 issue.
✘ Suite StorageSamplesDriver failed after 3.444 seconds with 1 issue.
✘ Test run with 1 test in 1 suite failed after 3.444 seconds with 1 issue.
```

That includes some debugging changes:

```
diff --git a/packages/swift-google-cloud-storage/Sources/GoogleCloudStorage/generated/StorageControl/Convert/GoogleLongRunning/Operation+Convert.swift b/packages/swift-google-cloud-storage/Sources/GoogleCloudStorage/generated/StorageControl/Convert/GoogleLongRunning/Operation+Convert.swift
[...]

Issue #588, opened 2026-09-01
[storage]: LRO helpers for storage control


coryan:
`StorageControlClient` is missing the LRO helpers, for example, `renameFolder()` gets a single overload:

https://github.com/googleapis/google-cloud-swift/blob/47c3e477f05d85ea0b929389e5dbd829c6b8b7b4/packages/swift-google-cloud-storage/Sources/GoogleCloudStorage/generated/StorageControl/StorageControlClient.swift#L430-L434

compare against `Workflows` which gets two overloads:

https://github.com/googleapis/google-cloud-swift/blob/47c3e477f05d85ea0b929389e5dbd829c6b8b7b4/generated/swift-google-cloud-workflows-v1/Sources/GoogleCloudWorkflowsV1/Workflows.swift#L89-L93

https://github.com/googleapis/google-cloud-swift/blob/47c3e477f05d85ea0b929389e5dbd829c6b8b7b4/generated/swift-google-cloud-workflows-v1/Sources/GoogleCloudWorkflowsV1/Workflows.swift#L100-L102

and some default implementations in the protocol:

https://github.com/googleapis/google-cloud-swift/blob/47c3e477f05d85ea0b929389e5dbd829c6b8b7b4/generated/swift-google-cloud-workflows-v1/Sources/GoogleCloudWorkflowsV1/Workflows.swift#L573-L599

PR #780, opened 2026-09-09
docs(storage): add storage_move_file sample

### LINKED ISSUE #772: Write the `storage_move_file` sample

coryan:
Fixes #772

### DIFF
--- a/Sources/StorageSamples/Objects/MoveFile.swift
+++ b/Sources/StorageSamples/Objects/MoveFile.swift
@@ -1,0 +1,54 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


Sources/StorageSamples/Objects/MoveFile.swift
```diff
@@ -0,0 +1,54 @@
...
+        $0.destinationBucket = "projects/_/buckets/\(destBucketId)"
+        $0.destinationName = destName
+        if !token.isEmpty {
+          $0.rewriteToken = token
+        }
+      },
+      options: .init()
+    )
+    token = response.rewriteToken
+  } while !response.done
```
chingor13, line 40: Is this really how the API works? Not an LRO?

chingor13 (approved):

Sources/StorageSamples/Objects/MoveFile.swift
coryan: ~They should be LROs, the generator has a bug #588 and (maybe) #594 .... I was planning to fix the examples after the generator produced good LRO helpers.~

Ah ... this is the rewrite thing.  Yes, these are not LROs.  We can add a helper function via some `extension` thing maybe post-GA. Tracking in #786

PR #787, opened 2026-09-09
feat(storage): convenience methods without options for StorageControlProtocol

nolanleastin:


packages/swift-google-cloud-storage/Tests/StorageControlClientTests.swift
```diff
@@ -72,4 +72,18 @@ import Testing
       _ = try StorageControlClient(options)
     }
   }
+
+  private func assertConvenienceSignatures(
```
coryan: You should update the storage samples to use these methods, and then you can remove the test I think.

nolanleastin: done

coryan (approved):
Maybe wait until #776 is merged and rebase?

Issue #785, opened 2026-09-09
Add request option to set the quota project


coryan:
Sometimes we need to override the default quota project, for example, in storage, when using "requester pays" the caller may need to set the quota project and the credentials override may be too coarse grained.

chingor13:
# Implementation Plan: Quota Project Support (`x-goog-user-project`)

This document outlines the Google Cloud specification for quota project attribution, how other client libraries (`google-cloud-rust`) implement it, and the proposed architecture for [`google-cloud-swift`](https://github.com/googleapis/google-cloud-swift/issues/785) and `librarian`.

---

## 1. Google Cloud Specification for Quota Projects

### Overview & Purpose
The **`x-goog-user-project`** HTTP header (and gRPC metadata key) designates the consumer Google Cloud project used for:
- **Quota attribution**: Charging API rate limits and quota usage against a specific project rather than a shared or credential-default project.
- **Billing attribution**: Required for client-based APIs and **Requester Pays** buckets in Cloud Storage, where the caller pays for data access/egress instead of the bucket owner.
- **API Enablement checks**: Service Control verifies that the target API is enabled on the designated quota project.

> [!IMPORTANT]
> The authenticated principal making the request must hold the `serviceusage.services.use` permission (e.g., via the **Service Usage Consumer** role `roles/serviceusage.serviceUsageConsumer`) on the project specified in `x-goog-user-project`.

### Precedence Hierarchy
Google Cloud client library standards specify a strict four-tier precedence hierarchy for determining the `x-goog-user-project` header (from highest to lowest):

1. **Per-Request Option (`RequestOptions.quotaProject` / `ReadObjectOptions.quotaProject` / `UploadOptions.quotaProject`)**:
   Explicitly set on a single RPC or data-plane operation. Must override any client-level or credential-level quota project.
2. **Client-Level Option (`ClientOptions.quotaProject`)**:
   Configured when instantiating a client; applies to all requests from that client unless overridden per request.
3. **Environment Variable (`GOOGLE_CLOUD_QUOTA_PROJECT`)**:
   Evaluated during Application Default Credentials (ADC) resolution (`ADC.resolve`).
4. **Credentials Configuration (`quota_project_id` in ADC JSON / `CredentialsConfiguration` / `workforcePoolUserProject`)**:
   Configured directly in credential key files (such as `authorized_user` JSON from `gcloud auth application-default set-quota-project`) or credential initializers.

### Header Suppression & Replacement Invariant
[...]

PR #791, opened 2026-09-09
fix(auth): normalize x-goog-user-project header key

chingor13:
fix(auth/user): store provided `quotaProjectID`, falling back to user
fix(auth/user): emit `x-goog-user-project` header if provided
fix(auth): Added quota_project_id decoding to ServiceAccountData and fallback resolution in ServiceAccountCredentials.

Towards #785


coryan (approved):
The outline from Gemini was really useful here. THanks.

PR #792, opened 2026-09-09
feat(gax): add per-request and per-client quota_project_id options

chingor13:
Towards #785

### DIFF
--- a/packages/swift-google-gax/Sources/GoogleCloudGax/ClientOptions.swift
+++ b/packages/swift-google-gax/Sources/GoogleCloudGax/ClientOptions.swift
@@ -60,8 +60,28 @@ public struct ClientOptions: Sendable {
 
   /// Overrides the quota project for all requests sent by the client.
   ///
+  /// By default, Google Cloud APIs attribute quota and billing usage to the project associated
+  /// with the credentials (for example, the project where a service account was created) or to
+  /// the project owning the accessed resource. Setting `quotaProject` instructs the service to
+  /// charge quota and billing to the specified project ID or project number instead.
+  ///
+  /// Common scenarios for setting a quota project include:
+  /// - Authenticating with user credentials (such as those created by
+  ///   `gcloud auth application-default login`), which are not inherently tied to a project.
+  /// - Accessing resources in another project (such as Cloud Storage Requester Pays buckets)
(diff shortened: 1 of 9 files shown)


packages/swift-google-gax/Sources/GoogleCloudGax/ClientOptions.swift
```diff
@@ -58,6 +58,12 @@ public struct ClientOptions: Sendable {
   /// [Application Default Credentials]: https://docs.cloud.google.com/docs/authentication/client-libraries
   public var credentials: Credentials? = nil
 
+  /// Overrides the quota project for all requests sent by the client.
+  ///
+  /// When set, the `x-goog-user-project` header is sent with this value on every request,
+  /// overriding any credential-level quota project unless overridden by ``RequestOptions/quotaProject``.
```
coryan: Maybe some background for a customer that is running into this for the first time:


```suggestion
  /// Overrides the quota project for all requests sent by the client.
  ///
  /// A [quota project] specifies the project used for quota and billing purposes. Commonly
  /// used when invoking APIs where the caller pays for the RPCs, or to charge a project
  /// other than the default for an RPC. Consult the API documentation for details about
  /// billing and quota.
  ///
  /// When set, the `x-goog-user-project` header is sent with this value on every request,
  /// overriding any credential-level quota project unless overridden by ``RequestOptions/quotaProject``.
  ///
  /// [quota project]:  https://docs.cloud.google.com/apis/docs/system-parameters#definitions
```

Don't take that as verbatim, the idea is that the reader may have no idea what "quota project" means and we want to leave some pointers to start their search .

packages/swift-google-gax/Sources/GoogleCloudGax/RequestOptions.swift
```diff
@@ -80,4 +80,10 @@ public struct RequestOptions: Sendable {
   ///
   /// Without an override, the request uses the polling backoff policy configured in the client.
   public var pollingBackoffPolicy: (any BackoffPolicy)? = nil
+
+  /// Overrides the quota project for this request.
+  ///
```
coryan, line 85: Ditto.

coryan (approved):

coryan (approved):

PR #816, opened 2026-09-10
feat(storage): add LRO polling helpers for storage

nolanleastin:


pkgs/swift-google-cloud-storage/Sources/GoogleCloudStorage/generated/StorageControl/StorageControlClient.swift
```diff
@@ -507,6 +569,30 @@ public final class StorageControlClient: StorageControlProtocol, Sendable {
...
+    withPolling: CreateAnywhereCacheRequest, options: GoogleCloudGax.RequestOptions
+  ) async throws -> any GoogleCloudGax.PollableOperation<AnywhereCache> {
+    let extractStatus = {
+      (op: GoogleLongRunning.Operation) throws
+        -> GoogleCloudGax._PollableOperationImpl<AnywhereCache>.State in
+      return try op._extractStatus(AnywhereCache.self)
+    }
+    let rawOp = try await self.createAnywhereCache(request: withPolling, options: options)
+    let initialState = try extractStatus(rawOp)
+    let poll = { () async throws -> GoogleCloudGax._PollableOperationImpl<AnywhereCache>.State in
```
coryan, line 583: Non-blocking: this assumes a successful conversion from Protobuf's `Any` to our `Any`.  As you are aware (#594) I noticed a problem, but could not repro in isolation.  An alternative might be to pass the expected types to `getOperation` such that it can convert them safely, but should be a separate change.

coryan (approved):

pkgs/swift-google-cloud-storage/Sources/GoogleCloudStorage/generated/StorageControl/StorageControlClient.swift
nolanleastin: Ok - will consider this next

PR #831, opened 2026-09-11
fix(wkt): `Empty` serialization into `Any`

### LINKED ISSUE #824: Fix serialization of `Empty` into `Any`
Gemini says:

#### 4. `google.protobuf.Empty` in `Any` Emits `"value": {}` (Explicit Spec Violation)
 
* **Spec Requirement**:

  > *"Note that `google.protobuf.Empty` is not considered to have any special JSON mapping; it is simply a normal message that has zero fields. This means the expected representation of an Empty packed into an Any is `{"@type": "type.googleapis.com/google.protobuf.Empty"}` and not `{"@type": "type.googleapis.com/google.protobuf.Empty", "value": {}}`."*
 
* **Current Implementation**:
  In [Empty.swift](file:///usr/local/google/home/coryan/google-cloud-swift/feat-wkt-protojson-int64-strings/pkgs/swift-google-wkt/Sources/GoogleCloudWKT/Empty.swift#L53-L55):
  ```swift
  public func _pack() throws -> Struct {
    return [`Any`.valueField: Value(object: [:])]
  }
  ```

  Packing an `Empty` into an `Any` produces:

  ```json
  {"@type": "type.googleapis.com/google.protobuf.Empty", "value": {}}
  ```
  which is the exact negative example called out in the specification.

coryan:
The ProtoJSON spec requires `Empty` to just omit the `value:` field. It
is the exception to the exception for well-known types.

Fixes #824

### DIFF
--- a/pkgs/swift-google-wkt/Tests/EmptyTests.swift
+++ b/pkgs/swift-google-wkt/Tests/EmptyTests.swift
@@ -95,7 +95,44 @@ import Testing
     let got = String(data: data, encoding: .utf8)!
 
     let want =
-      #"{"content":{"@type":"type.googleapis.com/google.protobuf.Empty","value":{}}}"#
+      #"{"content":{"@type":"type.googleapis.com/google.protobuf.Empty"}}"#
     #expect(got == want)
+  }
+
+  @Test("Roundtrip Empty through Any directly")
+  func emptyAnyDirectRoundtrip() throws {
+    let input = Empty()
+    let any = try `Any`(fromMessage: input)
+    #expect(any.typeUrl == "type.googleapis.com/google.protobuf.Empty")
+    let got = try Empty(fromAny: any)
+    #expect(got == input)
+  }
+
+  @Test("Unpack Empty from Any with invalid non-empty value")
+  func emptyAnyUnpackInvalidNonEmptyValue() throws {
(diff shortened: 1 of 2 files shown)


pkgs/swift-google-wkt/Tests/EmptyTests.swift
```diff
@@ -95,7 +95,44 @@ import Testing
...
+    let got = try Empty(fromAny: any)
+    #expect(got == input)
+  }
+
+  @Test("Unpack Empty from Any with invalid non-empty value")
+  func emptyAnyUnpackInvalidNonEmptyValue() throws {
+    let jsonString =
+      #"{"content":{"@type":"type.googleapis.com/google.protobuf.Empty","value":{"unexpected":"data"}}}"#
+    let data = Data(jsonString.utf8)
+    let decoder = JSONDecoder()
```
chingor13: Should we be using _ProtoJSONDecoder?

chingor13 (approved):

pkgs/swift-google-wkt/Tests/EmptyTests.swift
coryan: Yes, fixed here and the other tests in this file.   Created #837 to fix other files.

PR #872, opened 2026-09-14
fix(gax): rethrow non-RequestErrors in long-running operation polling

### LINKED ISSUE #725: Long-running operation polling misclassifies non-RequestError exceptions as UNIMPLEMENTED
Unconfirmed.  This is the result of a search with Gemini, may be a false positive.

In `_PollableOperationImpl.wait()`, when `pollOp()` throws an error, the catch block converts the error via `let requestError = (error as? RequestError) ?? .unimplemented`. If the thrown error is anything other than `RequestError`—including a Swift concurrency `CancellationError`, JSON/protobuf decoding errors, or underlying transport errors—it is unconditionally replaced with `RequestError.unimplemented` (gRPC status code 12).

Replacing arbitrary errors with `.unimplemented` prevents proper task cancellation handling and causes the error to be passed to `pollingPolicy.onError`. Because standard polling policies treat `UNIMPLEMENTED` as a non-retryable permanent failure, this masks the true underlying failure, drops cancellation signals, and throws a misleading "Unimplemented" error to the caller.

**References:**
- `packages/swift-google-gax/Sources/GoogleCloudGax/LongRunningOperations.swift:125-136`

chingor13:
In `_PollableOperationImpl.wait()`, when `pollOp()` threw an error that was not a `RequestError`, it was converted to `RequestError.unimplemented`. This masked the true failure, swallowed task `CancellationError`, and prevented cooperative cancellation.

Rethrow non-`RequestError`s directly, aligning behavior with `_RetryLoop`.

Fixes #725


pkgs/swift-google-gax/Sources/GoogleCloudGax/LongRunningOperations.swift
```diff
@@ -125,7 +125,9 @@ public final class _PollableOperationImpl<ResponseType>: PollableOperation {
       do {
         state = try await pollOp()
       } catch {
-        let requestError = (error as? RequestError) ?? .unimplemented
+        guard let requestError = error as? RequestError else {
+          throw error
+        }
```
coryan: optional: you could write 

```suggestion
        } catch let requestError as Request Error {
          .... your code here ...
        } catch {
          throw error
        }
```

coryan (approved):

Issue #866, opened 2026-09-13
Validate URL paths


coryan:
After support for multiple bindings, we need to implement  **go/client-libraries:rest-special-uri-chars** (Googlers-only). 

Basically we want to stop customers from using `..` or `.` in the URL paths because the request may not do what they think it will do.

- [ ] Implement Variable Validation (Prevent Dots)
   Before applying percent encoding or concatenating variables into the final URI string, the Swift runtime or generator must validate the extracted variable values.
   *   **For single-segment variables (`*`)**: Fail if the value is exactly `.` or `..`.
      *   *Error message:* `"Invalid value {. or ..} for {resource ID field name}."`
   *   **For multi-segment variables (`**`)**: Split the value by the path delimiter `/` and fail if **any segment** is exactly `.` or `..`.
      *   *Error message:* `"Value for { property name } must not contain segments that are exactly . or .. ."`

- [ ]  Apply Custom Percent Encoding
   Swift's default `.urlPathAllowed` `CharacterSet` may be too permissive or inconsistent with strict API guidelines. You must define a custom `CharacterSet` allowing only `[-_.~/0-9a-zA-Z]`.
   > **Note:** The `/` character must be kept unencoded in both `*` and `**` contexts. It won't exist in `*` (by definition), and it must stay unencoded in `**` to remain a valid path segment delimiter.

PR #883, opened 2026-09-15
feat(gax): validate and endcode URL paths

coryan:
This PR rejects resource names that can accidentally lead to path traversals and then querying or modifying the wrong object.

Towards #866


pkgs/swift-google-gax/Sources/GoogleCloudGax/RoutingMatcher.swift
```diff
@@ -122,4 +122,105 @@ public enum _RoutingMatcher {
     let end = haystack.index(haystack.startIndex, offsetBy: endOffset)
     return String(haystack[start..<end])
   }
+
+  /// The character set allowed in REST URI path templates per go/client-libraries:rest-special-uri-chars: `[-_.~/0-9a-zA-Z]`.
+  public static let restUriAllowedCharacterSet = CharacterSet(
```
chingor13: Does this need to be public? I guess this class is marked as `@_spi`, but it doesn't look like it needs to be public anyways.

chingor13 (approved):

pkgs/swift-google-gax/Sources/GoogleCloudGax/RoutingMatcher.swift
coryan: Good catch, thanks. Fixed.

PR #903, opened 2026-09-16
ci(docs): build documentation against the local packages

chingor13:
The documentation builds only put a hardcoded subset of the packages in this
repository into editable mode. Everything else resolved to the last published
version of the package, so a change that spans several packages fails to build.
For example, regenerating the libraries with support for unknown fields broke
the documentation builds, because the local `swift-google-cloud-storage` sources
were compiled against the published `swift-google-type`:

```
Date+Convert.swift:33:10: error: value of type 'Date' has no member '_unknownFields'
```

Use `ci/package-dependencies.sh` in both documentation builds, as `ci/test.sh`
and `ci/gcb/scripts/full.sh` already do. This overrides every local dependency
and restores them when the build finishes.

The list of local dependencies in that script was hand-maintained, which is the
same problem one level up: it goes stale the next time a package starts
depending on another package in this repository. Derive the list from the
manifests instead. The derived list matches the previous one, except for
`swift-google-iam-credentials-v1`, which no manifest references by URL, so
editing it was always a no-op.

Verified locally that `edit_package_dependencies "."` puts nine packages into
editable mode for the top-level package: `swift-google-api`, `-auth`,
`-cloud-location`, `-gax`, `-iam-v1`, `-longrunning`, `-rpc`, `-type` and
`-wkt`. The three the documentation builds previously missed (`-api`, `-rpc` and
`-type`) are the ones behind the failure above.


coryan (approved):
I suspect this force me to rethink / rebase #873, but your changes are more interesting / important.

PR #904, opened 2026-09-16
feat(generated): update generator with support for unknown keys

### LINKED ISSUE #447: sidekick: messages support unknown fields
The generated messages must support roundtrip serialization of unknown fields.

These fields must be stored in a private member variable, maybe as:

```swift
public struct MyMessage {
  private var _unknownFields: [String: GoogleCloudWkt.Value]
}
```

I think that means all messages will need to have a custom serialization and deserialization function.  We may want to split the serialization code to a separate file (`MyMessage+Additions.swift` ?).

chingor13:
Adds support for unknown keys

Fixes #447


coryan (approved):
I checked out the code on my laptop and reviewed the changes to handcrafted files there. LGTM.

PR #926, opened 2026-09-16
fix(storage): convert long-running operations that arrive from the wire

nolanleastin:
Regenerate storage to handle longrunning.Operation Any conversions


Sources/StorageSamples/RunFolderSamples.swift
```diff
@@ -54,9 +54,8 @@ public func runFolderSamples(
   try await getFolder(client: client, bucketId: id)
   print("running listFolders() sample")
   try await listFolders(client: client, bucketId: id)
-  // TODO(https://github.com/googleapis/google-cloud-swift/issues/594) - enable
```
coryan, line 57: There is a new set of samples that need to be uncommented. Search for the same TODO...

chingor13 (approved):

coryan (approved):

PR #936, opened 2026-09-17
doc: add skill and how-to for proposing releases

chingor13:
This may not be entirely correct, but we will edit as we go through an actual release.

Towards #914

Note: this adds a method for doing a synchronized release (e.g. we want to release 0.2.0 next for everything)


.agents/skills/create-release/SKILL.md
```diff
@@ -0,0 +1,232 @@
...
+  synchronized version bumps (such as dropping pre-release suffixes like `-preview`
+  or aligning all packages to a single version), regenerating affected packages,
+  validating changes, opening a pull request, and publishing split repositories.
+---
+
+# Create Release
+
+This skill guides the agent through creating a release for `google-cloud-swift`. It supports two release workflows:
+1. **Normal Incremental Release (`librarian bump`)**: Bumps versions for libraries modified since the last release tag using `librarian bump`, regenerates code, updates handwritten manifests if needed, and opens a release pull request.
+2. **Synchronized Version Bump**: Aligns all libraries across the repository to a specific target version (e.g., `0.2.0`) or transitions pre-release identifiers (such as dropping `-preview`).
```
coryan, line 15: This will be rare after GA, you should consider a separate skill or skills.

doc/contributor/howto-guide-releases.md
```diff
@@ -0,0 +1,226 @@
...
+**synchronized version bumps** (such as aligning all packages to a specific
+version or dropping pre-release suffixes like `-preview`).
+
+## Overview
+
+Releases in `google-cloud-swift` follow a two-phase workflow:
+
+1. **Release Preparation (Pull Request)**:
+   - Update library versions in [`librarian.yaml`](../../librarian.yaml).
+   - Update dependency version constraints in handwritten `Package.swift` manifests if shared dependencies were bumped.
```
coryan, line 14: FWIW, librarian for Rust knows how to do this (it edits a partially handwritten `Cargo.toml` file), we may want to consider a similar feature for Swift.

doc/contributor/howto-guide-releases.md
```diff
@@ -0,0 +1,226 @@
...
+
+```bash
+git checkout main
+git pull --ff-only upstream main
+git checkout -b chore-release-$(date +%Y-%m-%d)
+```
+
+### Step 2: Bump Library Versions
+
+Retrieve the configured `librarian` version and run `librarian bump`:
```
coryan, line 50: You should warn folks that `librarian bump` may change nothing... it only bumps once until the next release. In principle, we could call it on each PR and it would only affect the generated code once.

.agents/skills/create-release/SKILL.md
coryan: I am not being clever: we can remove this section after GA ... that would work too

coryan (approved):

doc/contributor/howto-guide-releases.md
chingor13: I **think** this is already handled, because we run the generator after bumping the versions (to update the version # in generated code and regenerate all the Package.swift files)

chingor13: Oh, this is for the hand-crafted Package.swift files

doc/contributor/howto-guide-releases.md
chingor13: Done

Issue #932, opened 2026-09-17
release 0.2.0 version


chingor13:
* drop the `-preview` in the version
* synchronize the versions for now

chingor13:
This is done

Issue #846, opened 2026-09-11
cleanup dependencies in guide


chingor13:
if the packages used in guide/ are in split repos then we can start fixing the user guide. Just saying.

https://github.com/googleapis/google-cloud-swift/pull/845#pullrequestreview-5183597514

PR #945, opened 2026-09-18
chore: release 0.2.0

### LINKED ISSUE #942: Cannot bump multiple versions at the same time
## Why swift package edit creates a chicken-and-egg problem on release PRs
Even when called in a loop one by one (as 
ci/package-dependencies.sh
 originally did), swift package edit fails when multiple interdependent packages are bumped to an unpublished version (0.2.0):

1. Suppose a package depends on both swift-google-wkt (from: "0.2.0") and swift-google-gax (from: "0.2.0").
2. When you run the first command:
    ```bash
    swift package edit --path "${REPO_ROOT}/pkgs/swift-google-wkt" swift-google-wkt
    ```
    SwiftPM must first load the package graph and resolve all other dependencies in Package.swift before applying the edit.
3. Because swift-google-gax is not in edit mode yet, SwiftPM queries GitHub (https://github.com/googleapis/swift-google-gax) for a 0.2.0 tag, fails to find one, and aborts before putting swift-google-wkt into editable mode.
4. Every subsequent swift package edit call in the loop fails for the exact same reason.

(Note: swift package config set-mirror has a similar limitation—it redirects URLs, but SwiftPM still treats from: "0.2.0" as a git tag requirement on the mirrored repository rather than a local path dependency.)

chingor13:
## Summary

This PR updates all library versions to `0.2.0`, regenerates all client libraries using the updated `librarian` generator (https://github.com/googleapis/librarian/pull/7615), and updates local dependency resolution across CI and package manifests.

### Commits
1. `chore: bump versions to 0.2.0 in librarian.yaml` — Updates `default_version` and library `version` entries in `librarian.yaml` to `0.2.0`.
2. `chore: regenerate libraries for 0.2.0` — Re-runs `librarian generate --all` using https://github.com/googleapis/librarian/pull/7615, updating generated `Package.swift` manifests to use `localOrRemotePackage` and bumping version constants to `0.2.0`.
3. `build: support local dependency resolution via GOOGLE_CLOUD_SWIFT_LOCAL_DEPS` — Updates `Package.swift`, `Package.resolved`, handwritten package manifests, and CI scripts (`ci/package-dependencies.sh`, `ci/gcb/scripts/fetch.sh`, `.github/workflows/build.yml`) to resolve local monorepo packages via `GOOGLE_CLOUD_SWIFT_LOCAL_DEPS`.

Towards #932
Fixes #942


ci/package-dependencies.sh
```diff
@@ -14,60 +14,49 @@
...
+    else
+        _CREATED_RESOLVED+=("${dir}/Package.resolved")
+    fi
+
+    # SwiftPM requires automatic resolution when dependencies are overridden.
     if [[ -n "${flags+x}" ]]; then
         local filtered_flags=()
         local had_flag=false
-        for f in "${flags[@]}"; do
+        for f in ${flags[@]+"${flags[@]}"}; do
```
coryan, line 59: The `${flags[@]+ ...` is to avoid problems if flags is unset, right? But you already checked that a couple of lines before>

ci/package-dependencies.sh
```diff
@@ -76,7 +65,7 @@ edit_package_dependencies() {
         done
         if [[ "${had_flag}" == true ]]; then
             _REMOVED_DISABLE_RESOLUTION+=("${dir}")
-            flags=("${filtered_flags[@]}")
+            flags=(${filtered_flags[@]+"${filtered_flags[@]}"})
```
coryan, line 68: Again, overkill

ci/package-dependencies.sh
```diff
@@ -85,50 +74,57 @@ restore_package_dependencies() {
     local dir="$1"
     local clean_dir="${dir#./}"
     [[ -z "${clean_dir}" ]] && clean_dir="."
-    local scratch_args=()
-    if [[ -n "${flags+x}" ]]; then
-        for ((i=0; i<${#flags[@]}; i++)); do
-            if [[ "${flags[i]}" == "--scratch-path" && $((i+1)) -lt ${#flags[@]} ]]; then
-                scratch_args=("--scratch-path" "${flags[i+1]}")
+
+    if [[ -f "${dir}/Package.resolved.ci-bak" ]]; then
```
coryan, line 78: Food for thought, can we just not restore the Package.resolved file in CI builds? We just drop the thing on the floor?

ci/package-dependencies.sh
```diff
@@ -85,50 +74,57 @@ restore_package_dependencies() {
...
+    unset GOOGLE_CLOUD_SWIFT_LOCAL_DEPS
+
+    for res in ${_MODIFIED_RESOLVED[@]+"${_MODIFIED_RESOLVED[@]}"}; do
+        if [[ -f "${res}.ci-bak" ]]; then
+            mv "${res}.ci-bak" "${res}"
+        fi
+    done
+    _MODIFIED_RESOLVED=()
+
+    for res in ${_CREATED_RESOLVED[@]+"${_CREATED_RESOLVED[@]}"}; do
```
coryan, line 124: Methinks your agent got trigger happy about `[@]+`

guide/Package.swift
```diff
@@ -25,8 +25,16 @@ let package = Package(
...
+      from: "0.2.0"
+    ),
+    localOrRemotePackage(
+      url: "https://github.com/googleapis/swift-google-gax",
+      path: "pkgs/swift-google-gax",
+      from: "0.2.0"
+    ),
     .package(path: "../generated/swift-google-cloud-secretmanager-v1"),
     .package(path: "../generated/swift-google-cloud-language-v2"),
     .package(path: "../generated/swift-google-cloud-workflows-v1"),
```
coryan, line 40: Should we change these too?

ci/package-dependencies.sh
coryan: You can clean up these things in a follow up PR btw.

coryan (approved):

guide/Package.swift
chingor13: I think we can follow up in #846

ci/package-dependencies.sh
chingor13: Ack.

